PR #2, opened 2015-05-05
Allow items passed to preCache to be Promises for URLs as well as just URLs

wibblymat:
R: @jeffposnick @addyosmani


lib/sw-toolbox.js
```diff
@@ -32,7 +32,9 @@ self.addEventListener('install', function(event) {
   helpers.debug('preCache list: ' + (options.preCacheItems.join(', ') || '(none)'));
   event.waitUntil(
     helpers.openCache({cacheName: inactiveCache}).then(function(cache) {
-      return cache.addAll(options.preCacheItems);
+      return Promise.all(options.preCacheItems).then(function(items) {
```
jeffposnick: I believe you can omit the anonymous function and just write `return Promise.all(options.preCacheItems).then(cache.addAll);`

addyosmani: :+1:

wibblymat: Needed a `bind`, but done.

jeffposnick:
:+1:

PR #5, opened 2015-05-07
Added some basic tests

wibblymat:
R: @jeffposnick @jakearchibald

I never found a satisfactory way of running the tests in the worker, but hopefully this will work.

Doesn't currently cover all features, but some is better than none.

### DIFF
--- a/tests/index.html
+++ b/tests/index.html
@@ -1,0 +1,13 @@
+<!doctype html>
+<html>
+  <head>
+    <title>Service Worker Toolbox Tests</title>
+    <link rel="stylesheet" href="//code.jquery.com/qunit/qunit-1.17.0.css"/>
+  </head>
+  <body>
+    <div id="qunit"></div>
+    <div id="qunit-fixture"></div>
+    <script src="//code.jquery.com/qunit/qunit-1.17.0.js"></script>
+    <script src="tests.js"></script>
+  </body>
+</html>
--- a/tests/index.html
+++ b/tests/index.html
@@ -2,12 +2,12 @@
 <html>
   <head>
     <title>Service Worker Toolbox Tests</title>
-    <link rel="stylesheet" href="//code.jquery.com/qunit/qunit-1.17.0.css"/>
+    <link rel="stylesheet" href="../node_modules/qunitjs/qunit/qunit.css"/>
   </head>
   <body>
     <div id="qunit"></div>
     <div id="qunit-fixture"></div>
-    <script src="//code.jquery.com/qunit/qunit-1.17.0.js"></script>
+    <script src="../node_modules/qunitjs/qunit/qunit.js"></script>
     <script src="tests.js"></script>
   </body>
 </html>
--- a/tests/tests.js
+++ b/tests/tests.js
@@ -1,0 +1,79 @@
+'use strict';
+
+navigator.serviceWorker.register('service-worker.js', { scope: './' });
+
+var checkValue = function(url, value, assert, method) {
+  var done = assert.async();
+  method = method || 'get';
+  return fetch(url, {method: method}).then(function(response) {
+    return response.text();
+  }).then(function(text) {
+    assert.equal(text, value);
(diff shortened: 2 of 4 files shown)


tests/tests.js
```diff
@@ -0,0 +1,79 @@
...
+  QUnit.test('First declared route wins', function(assert) {
+    // Matches both routes
+    checkValue('multiple/match/anything.html', '1', assert);
+
+    // Only matches second route
+    checkValue('multiple/match/anything', '2', assert);
+  });
+
+  QUnit.test('Caching', function(assert) {
+    var date = +(new Date());
```
jeffposnick: `Date.now()` would be clearer.

tests/tests.js
```diff
@@ -0,0 +1,79 @@
...
+    var url = 'cache/' + date;
+    var done = assert.async();
+    // Empty cache
+    fetch(url).then(function(response) {
```
jeffposnick: I'm having a hard time following the flow here, and understanding what each step is performing.

Could you a) break out the rejection handlers into a separate `.catch()` if possible (that tends to make things more readable), and b) add some comments?

tests/index.html
```diff
@@ -0,0 +1,13 @@
+<!doctype html>
+<html>
+  <head>
+    <title>Service Worker Toolbox Tests</title>
+    <link rel="stylesheet" href="//code.jquery.com/qunit/qunit-1.17.0.css"/>
```
jeffposnick: Can you pull in `qunit` locally via a `devDevependency` in your `package.json` and use that? I'm sure `code.jquery.com` isn't going away anytime soon, but I don't see the benefit of depending on something remote.

jeffposnick:
Thanks for writing tests! They look good in general.

My main piece of feedback is that the tests would be more useful if they could be automated via some sort of runner, which was then tied in to the release process. I've had good luck using `selenium` to test the `<platinum-sw>` elements in supported browsers via the [`web-component-tester`](https://github.com/Polymer/web-component-tester) framework, and maybe that's something you could borrow (without all the web component-y bits).

I don't think that needs to block merging this PR, but automation could be a future enhancement.

wibblymat:
Definitely agree about automation! I'd love to run this through Travis. I've opened a separate issue for that (#15).

Will resolve the other issues.

addyosmani:
> My main piece of feedback is that the tests would be more useful if they could be automated via some sort of runner, which was then tied in to the release process.

@jeffposnick are you familiar with our Polymer CI runner? http://polymerlabs.github.io/ci-runner/#!/PolymerElements/iron-a11y-keys/1ab9e69aff38a8712dffdfe6749a6d83fcab9e84

One of the benefits is that it's able to run your WCT tests through CI each time a PR comes in, which can be handy for making sure your tests aren't breaking with new patches coming through.

jeffposnick:
@addyosmani Yep, I am! It would be awesome if Mat could move towards something like that, though WCT as-is might not be an option, since his tests are QUnit-based and he doesn't have a web component.

I'm not entirely happy with the [`sw-precache` approach](https://github.com/GoogleChrome/sw-precache/blob/master/test/test.js), but it's another option—I use `mocha` to test the behavior of various JS helper methods, but don't test the actual service worker in-browser behavior.

wibblymat:
@jeffposnick PTAL

jeffposnick:
Not sure if you're opposed to pulling in QUnit as a local dependency or if you just missed that comment. If you'd rather continue using the CDN version, that's fine too, and feel free to merge.

addyosmani:
My two cents are a local dep for QUnit would make sense here, but if the CDN version is preferable and we intend to move over to something else long term anyway, that also sgtm. 

PR itself LGTM.

wibblymat:
Sorry, actually missed the qunit comment first time around. Have brought it in locally. Will merge, but feel free to check over the latest commit :)

addyosmani:
:+1:

PR #17, opened 2015-06-05
RegExp objects for the same pattern are not equal

wibblymat:
Dumb mistake when I moved to using Maps meant that every route creates a seperate `origin` Map, even if the origin is actually the same. Using the string source of the RegExps (as I had originally) should solve it.

Handily this was caught by failing tests in the tests PR!

R: @addyosmani @jeffposnick


wibblymat:
I'm confident of this change and consider it to be urgent as things are basically broken without out it.

Please feel free to review post-merge!

addyosmani:
:+1:

jeffposnick:
One question I had regarding the use of `RegExp.source`: as [per MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/source), it won't include any of the RegExp flags, and I'm wondering whether ignoring those could have any real-world effect. I'm guessing not, but wanted to double-check with you.

One scenario that may or may not be an issue is, e.g., the developer passing in a flag of `i` along with a origin pattern that contains uppercase characters, and expecting there to be a insensitive match done? Are you normalizing the case of the origins anywhere downstream?

wibblymat:
Should have responded to this ages ago.

@jeffposnick: I hadn't actually considered the flags situation. I believe that case sensitivity is the only real gotcha here. I'm not sure what to do about it. Tempted to just wait and see.

PR #23, opened 2015-07-28
Fixed a bug in the `fastest` strategy.

wibblymat:
So, turns out that `fastest` has a horrible bug. I will tag a new release as soon as this is fixed.

`cacheOnly` will resolve with `undefined` if the resource isn't in the cache, rather than rejecting. We need to account for this and assume that resolving with a non-Response is a failure.

R: @jeffposnick


jeffposnick:
:+1:

PR #25, opened 2015-07-30
Flatten array of items to precache. Fixes #24

### LINKED ISSUE #24: Flatten precache() results before passing to cache.addAll()
https://github.com/GoogleChrome/sw-toolbox/blob/79def66583319e2da08cb962ed29cc0af8a2a0c9/lib/sw-toolbox.js#L35 is currently `return Promise.all(options.preCacheItems).then(cache.addAll.bind(cache));`

`cache.addAll()` takes in a one-dimensional array of either URLs or `Request` objects, but `Promise.all()` might resolve to a multi-dimensional array if the underlying promises in `options.preCacheItems` each resolve to an array.

`sw-toolbox` should flatten the value that `Promise.all()` resolves with before passing it along to `cache.addAll()`.

wibblymat:
I even wrote tests :)

R: @jeffposnick


jeffposnick:
:+1:, and `<platinum-sw-cache>` is happy when I test against the change.

Issue #9, opened 2015-05-18
Create a good demo


wibblymat:

jeffposnick:
Thinking about this some more, it would probably be a good idea to come up with a set of focused "recipes" for common usage patterns, rather than just a single demo. We can track them as subitems in this bug if you're amenable to that idea.

addyosmani:
I'm :+1: for the recipes pattern. We're finding increasing value for this in other projects and it's a little less work to update a recipe than a fully blown sample when things change (citation needed) :)

MoOx:
Any place(s) where I can find some examples? I am very new to sw and would like to see some best practices using this library (which looks very nice!).

MoOx:
Maybe I should look sw-precache instead. Can't find examples :/

andreban:
Seems like a simple demo on how to use the library would be very helpful. I tried SW with Vanilla JS and I can tell that sw-toolbox makes life much simpler. The hardest part for sw-toolbox was figuring out how everything fits together.

PR #34, opened 2015-09-21
networkFirst timeout

### LINKED ISSUE #12: Set a short timeout for networkFirst
And make it configurable.

jeffposnick:
R: @wibblymat 

Closes #12, although it doesn't enforce the network timeout by default.

I put together a super-simple page in `demo/index.html` that registers a service worker, mainly to play around with in DevTools. It can be fleshed out a bit more in the future, which is tracked in #9.

None of the existing tests that exercise `networkFirst` break, but I didn't add any new ones—if you have any ideas for simulating a slow network for just one test case, I'm all ears.

There are also miscellaneous changes to the docs—new docs for `networkTimeoutSeconds`, updates to a few copyright years, and correcting the documentation for the `cacheName` option.


LICENSE
```diff
@@ -186,7 +186,7 @@ Apache License
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright ${copyright_attribution}
+   Copyright 2015 Google, Inc.
```
wibblymat: I "fixed" this too, early on, but if you read the text this is NOT the copyright statement for the project, but rather the template for how to apply the license to other projects.

README.md
```diff
@@ -141,7 +147,7 @@ Patches are encouraged, and may be submitted by forking this project and submitt
 
 ## License
 
-Copyright 2014 Google, Inc.
+Copyright 2015 Google, Inc.
```
wibblymat: We should confirm with OSPO, but I've heard that we might actually be able to remove the year from copyright statements, which helps avoid doing an annual copyright PR.

README.md
```diff
@@ -105,7 +105,13 @@ Handle the request by trying to fetch the URL from the network. If the fetch fai
 ### Global Options
 Any method that accepts an `options` object will accept a boolean option of `debug`. When true this causes Service Worker Toolbox to output verbose log messages to the worker's console.
 
-Most methods that involve a cache (`toolbox.cache`, `toolbox.uncache`, `toolbox.fastest`, `toolbox.cacheFirst`, `toolbox.cacheOnly`, `toolbox.networkFirst`) accept an option called `cache`, which is the **name** of the [Cache](https://slightlyoff.github.io/ServiceWorker/spec/service_worker/#cache) that should be used. If not specifed Service Worker Toolbox will use a default cache.
+Most methods that involve a cache (`toolbox.cache`, `toolbox.uncache`, `toolbox.fastest`, `toolbox.cacheFirst`, `toolbox.cacheOnly`, `toolbox.networkFirst`) accept an option called `cacheName`, which is the **name** of the [Cache](https://slightlyoff.github.io/ServiceWorker/spec/service_worker/#cache) that should be used. If not specified, Service Worker Toolbox will use a default cache.
+
+The `networkFirst` strategy supports a timeout, specified in seconds via the `networkTimeoutSeconds`
```
wibblymat: I know that I would have just called it `timeout`, but I also know that I'm lazy and fond of shortcuts. I leave it to your judgement :)

wibblymat:
LGTM, resolve the remaining two nits or not at your discretion :)

README.md
jeffposnick: I don't think anyone will mind the verbosity. If there ends up being some other sort of configurable timeout in the future (not outside the realm of possibility), using a more descriptive identifier here will end up paying off.

README.md
jeffposnick: We can track that/follow up outside this PR.

PR #37, opened 2015-09-22
Adds cache expiration options

### LINKED ISSUE #18: Support for cache expiration
Brainstorming a bit: it would be useful if `sw-toolbox` had built-in support for cache expiration. It might take the form of associating LRU or oldest-first expiration policies with a specific named `Cache` object, along with a maximum number of entries, and then enforcing those policies via the built-in methods that store new entries. IndexedDB could be used to aid in bookkeeping.

`sw-toolbox` currently makes some assumptions about there only being a single named cached that your app works with globally, so relaxing some of those assumptions might be part of this work.

jeffposnick:
R: @wibblymat @jakearchibald @addyosmani (wouldn't mind getting several pairs of eyes on this)

This adds a new option, `maxCacheEntries`, along with some logic to implement the LRU expiration via IndexedDB.

It _seems_ to work, and it tries not to be terribly inefficient by
- running the cache expiration logic outside of the `FetchEvent.respondWith()`, so that it won't block the response
- reusing the `Cache` when possible, and the `IndexedDB` connection for the lifetime of the service worker, instead of closing and reopening them 

The dummy page in `demo/` was updated to trigger LRU expirations and the associated debug logging to the `console`.

Feedback and suggestions about alternate implementations are welcome.

Closes #18


lib/helpers.js
```diff
@@ -35,12 +36,22 @@ function openCache(options) {
...
 
-    // Only cache GET requests with successful responses
+  return fetch(request.clone()).then(function(response) {
+    // Only cache GET requests with successful responses.
+    // Since this is not part of the promise chain, it will be done asynchronously and will not
+    // block the response from being returned to the page.
     if (request.method === 'GET' && successResponses.test(response.status)) {
-      openCache(options).then(function(cache) {
-        cache.put(request, response);
+      openCache({cacheName: cacheName}).then(function(cache) {
```
wibblymat: Why not just pass `options` to `openCache` as it was before?

wibblymat:
In the process of resolving merge conflicts on the eslint branch I have ended up doing a force push of a rebased tree to master. I suck at git, apparently!

The end result is that this PR is a bit messed up. After merging everything I've got it so that the diff looks right, but the list of commits is wrong. It probably needs squashing, but I'm wary of doing further rebasing myself.

wibblymat:
Looks like I also confused @googlebot

lib/simple-db.js
```diff
@@ -0,0 +1,173 @@
+/**
```
wibblymat: This file doesn't pass the new ESLint checks. Not sure if we should be spinning this out into a separate module or fixing it up here.

lib/helpers.js
jeffposnick: I'm already doing the `cacheName = options.cacheName || globalOptions.cacheName` earlier in the code, so I figured it wasn't necessary to trigger that logic again inside the `openCache()` method. But I can restore it back to the original if you think it's cleaner.

(Refactoring all the options logic to use a helper like [`_.defaults()`](https://lodash.com/docs#defaults) might be nice in a future PR.)

jeffposnick:
I can squash the commits when we're happy with the code, prior to merging.

lib/simple-db.js
jeffposnick: I'll fix it up here. I've used this code elsewhere and have some confidence that it works as expected, so I'd rather stick with it rather than pulling in an IDB Promise wrapper from `npm`. If we do refactor this out to be a standalone library it should be a quick PR in the future.

jeffposnick:
PTAL.

I've also added the first "recipe" (see #9) to show off how `maxCacheEntries` can be used to maintain an independent cache of dynamic, large-ish, resources: YouTube thumbnails.

lib/simple-db.js
jeffposnick: It should pass the checks now.

addyosmani:
Logic changes LGTM. Will defer to @wibblymat on any remaining ESLint issues.

lib/helpers.js
```diff
@@ -47,6 +57,72 @@ function fetchAndCache(request, options) {
...
+function idbPromise() {
+  if (staticIdb) {
+    return Promise.resolve(staticIdb);
+  }
+
+  return simpleDB.open('sw-toolbox-lru-order').then(function(idb) {
+    staticIdb = idb;
+    return staticIdb;
+  });
+}
```
wibblymat: Since we only ever return a promise, we might as well store the promise rather than the value.

Alternative:
```javascript
function idbPromise() {
  if (staticIdb) {
    return staticIdb;
  }

  return staticIdb = simpleDB.open('sw-toolbox-lru-order');
}
```

lib/helpers.js
```diff
@@ -47,6 +57,72 @@ function fetchAndCache(request, options) {
...
+  var cacheExpiration = cacheExpirationPromiseFactory.bind(null, request, cache, options);
+
+  if (cacheExpirationPromiseChain) {
+    cacheExpirationPromiseChain = cacheExpirationPromiseChain.then(cacheExpiration);
+  } else {
+    cacheExpirationPromiseChain = cacheExpiration();
+  }
+}
+
+function cacheExpirationPromiseFactory(request, cache, options) {
```
wibblymat: Possible simplification:
- store a 'last-used' timestamp with each URL
- do an update-or-insert for the URL we just cached to set the timestamp to now
- use IDBIndex.count() to check whether we are over quota
- use a cursor on the index to get the correct number of least-recently-used items and delete them.

Actually, now that I write it out I'm not sure if it really is simpler. It just seems that fetching the whole data set as an array and manipulating that is inefficient somehow.

Bonus: Using a timestamp with an index rather than insert order means that we get time-based cache expiration almost for free.

recipes/max-cache-entries/index.html
```diff
@@ -0,0 +1,56 @@
...
+    <link rel="stylesheet" href="../common.css">
+    <link rel="stylesheet" href="styles.css">
+  </head>
+
+  <body>
+    <h1>Max Cache Entries Demo</h1>
+
+    <h3>Background</h3>
+    <p>
+      This configuration in
```
wibblymat: Wordsmithing:

*This example demonstrates using the `maxCacheEntries` option to restrict the number of items that will be stored in the cache. It uses a dedicated cache to hold YouTube video thumbnails. That dedicated cache will...* [continue as before]

wibblymat:
Some weirdness occurs to me. We are trying to enforce a restriction on a *cache*, but really what we wanted was to enforce the restriction on the *route*. That is, there should only be 10 images in the cache.

Let's say that someone does:
```javascript
toolbox.get('/images/:name.jpg', toolbox.cacheFirst, {maxCacheEntries: 10});
toolbox.get('/js/(.*)', toolbox.cacheFirst);
```

What might they expect to happen? I think most people would assume that this will cache all of their scripts but only up to 10 images. But actually, every time they fetch an image, some number of cached scripts could be discarded.

Some possibilities:
- Set the options on the **cache**, something like `toolbox.cacheOptions('myCache', {maxCacheEntries: 10}`. Then every route that specifies this cache will use the same options.
- Make the max size work per route. So you have to store some sort of ID for the route in IDB.

wibblymat:
Actually, ignore me - if you don't set `maxCacheEntries` on a particular route the code is never triggered, and the items getting added aren't included in the IDB of eviction candidates. So the first example I gave works as you'd expect.

Still some possible weirdness if you have:
```javascript
toolbox.get('/images/:name.jpg', toolbox.cacheFirst, {maxCacheEntries: 10});
toolbox.get('/videos/video.mp4', toolbox.cacheFirst, {maxCacheEntries: 20});
```
because the IDB table is per cache.

jeffposnick:
@wibblymat, what do you think about requiring that setting the max entries always be used in conjunction with setting the name? If we ask developers to explicitly name the cache that they're imposing the size restriction on, then they'll tend to choose different names for different purposes. If they choose the same cache name but different max entries for different routes then the "weirdness" would still be there, but it is something that they effectively opted into.

The call could look like:

```js
toolbox.get('/(.*).jpg', toolbox.networkFirst, {
  cache: {
    name: 'my-image-cache',
    maxEntries: 10
  }
});
````

We could still support

```js
toolbox.get('/(.*).jpg', toolbox.networkFirst, {
  cacheName: 'my-image-cache'
});
```

as well as

```js
toolbox.get('/(.*).jpg', toolbox.networkFirst, {
  cache: {
    name: 'my-image-cache'
  }
});
````

for developers who want to use a custom cache name but don't want cache expiration, but 

```js
toolbox.get('/(.*).jpg', toolbox.networkFirst, {
  cache: {
    maxEntries: 10
  }
});
````

would just log an error about needing to set the cache name.

File: lib/helpers.js (hunk 3 above)
Comment (jeffposnick) [inline lib/helpers.js]: Done.

File: lib/helpers.js (hunk 4 above)
Comment (jeffposnick) [inline lib/helpers.js]: Sigh. I had wanted to avoid having to use IndexedDB as, like, a DB with indexes and all, but I guess you're right about it being more efficient (though not simpler).

I'll take a look at making those modifications, though it might be abandoning the `SimpleDB` wrapper and going directly against the asynchronous API.

**Comment by jeffposnick** [comment]
I just pushed a new commit with the refactoring to use `IndexedDB` to store url + timestamp pairs instead of just keeping track of the ordering in an `Array`. I also added in time-based expiration, since, as @wibblymat mentioned, it's pretty straightforward with that `IndexedDB` approach.

I didn't change the cache options as described in https://github.com/GoogleChrome/sw-toolbox/pull/37#issuecomment-143257834 yet, though. I'll sync up with @wibblymat on that before proceeding.

File: lib/helpers.js (hunk 4 above)
Comment (jeffposnick) [inline lib/helpers.js]: Okay, I pushed a new commit that uses `IndexedDB` directly.

**Comment by jeffposnick** [comment]
Okay! I've refactored the way that cache configuration is specified and update the docs for the options to make things a little clearer. A further docs touch-up might be forthcoming, but I don't want to add any more to what's already a very large PR.

If you're happy with this, @wibblymat, hopefully we can merge. There's a breaking change to the existing `toolbox.options.cacheName` global, so it will have to be part of an eventual `3.0.0` release.

File: README.md [hunk 6]
```diff
@@ -103,15 +103,50 @@ Handle the request by trying to fetch the URL from the network. If the fetch fai
...
+`networkTimeoutSeconds` is not set, the browser's native networking timeout logic applies.
+_Default_: `null`
+
+#### cache [Object]
+Various properties of `cache` control the behavior of the default cache when set via
+`toolbox.options.cache`, or the cache used by a specific request handler.
+
+#### cache.name [String]
+The name of [`Cache`](https://slightlyoff.github.io/ServiceWorker/spec/service_worker/index.html#cache)
+used to store [`Response`](https://fetch.spec.whatwg.org/#response-class)s. Using a uniquely name
```
wibblymat: typo: uniquely -> unique

wibblymat:
Just nits - LGTM!

jeffposnick:
Cool! I'm going to give rebasing this PR a shot, but if that gets too gnarly, then I might just merge.

jeffposnick:
Merging as-is, as `git rebase -i origin/master` seems to just take me through the same procession of merge conflicts that you already fought through once.

PR #39, opened 2015-09-25
Fixes #21. The default route should only apply to GET requests

### LINKED ISSUE #21: Returning cached resource based on route without considering method
I'm using sw-toolbox through Polymer's `platinum-sw-cache` and `iron-ajax` and it's come up with a bit of strange behaviour explained below:
```
GET /api/resource --> 200 {"data": [{"id": 1}, {"id":2}]
POST /api/resource --> Server sends 422 {"errors": [{"description": "can't be blank"}]}, SW intercepts and returns 200 {"data": [{"id": 1}, {"id":2}] `Status Code:200 OK (from ServiceWorker)`
```
However, When inspecting the service worker, it shows 2 Requests:
```
Request Method:OPTIONS
Status Code:200 OK
```
```
Request Method:POST
Status Code:422 Unprocessable Entity
```
and `POST https://localhost:4000/api/resource 422 (Unprocessable Entity)` in the console

TBH, I don't really expect anything but GET to be cached, which may the case, but it definitely shouldn't replace my POST's 422 and body with a 200 OK and the GET body.

wibblymat:
R: @jeffposnick 

I can't work out if this is controversial or not. Since you can't cache non-GET requests, it seems safe to me.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ toolbox.router.get(':foo/index.html', function(request, values) {
 // For requests to other origins, specify the origin as an option
 toolbox.router.post('/(.*)', apiHandler, {origin: 'https://api.example.com'});
 
-// Provide a default handler
+// Provide a default handler for GET requests
 toolbox.router.default = myDefaultRequestHandler;
 
 // You can provide a list of resources which will be cached at service worker install time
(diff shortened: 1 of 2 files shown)


jeffposnick:
That looks like a safe change to make. I can't imagine that anyone would end up defining a default route that worked well for non-`GET` requests.

Could you please make a change to [the docs](https://github.com/GoogleChrome/sw-toolbox/blob/5200ff6d76f738592efe9a5c61875c9b68a0e89d/README.md#basic-usage) to switch `// Provide a default handler` -> `// Provide a default handler for GET requests.`

Feel free to merge once the docs are updated.

PR #40, opened 2015-09-25
Don't exclude the library itself from npm

### LINKED ISSUE #31: Want to be able to use the Npmjs package as a Node library in builds
First, thanks for this tool, it looks really great and promises to save tons of boilerplate.

I'm trying to use it as a component library in my own service worker build. I'm bundling the service worker instead of using importScripts for each module - Want one trip (no HTTP/2).

I'm using the sw-toolbox npm package 2.0.4, and I'm having difficulties:
  1. Uglify chokes on the pre-built source code.
  2. No 'main' in package.json, so require('sw-toolbox/sw-toolbox') is weird.

The raw error (Uglifying the built code):
```
WARNING in ./~/sw-toolbox/sw-toolbox.js
Critical dependencies:
17:438-445 This seems to be a pre-built javascript file. Though this is possible, it's not recommended. Try to require the original source to get better results.
 @ ./~/sw-toolbox/sw-toolbox.js 17:438-445

ERROR in sw.84e0fb9382841b26f9da.min.js from UglifyJs
Unexpected token name «of», expected punc «;» [./~/sw-toolbox/sw-toolbox.js:50,0]
```

The npmjs package only has the pre-built source code, not the library code.
I know I can just use the git repo, but prefer npmjs and node package management semantics througout the project.

Any reason not to just publish the library source code in the npm package for a more "node aligned" experience?

wibblymat:
Allows other developers to manage the build themselves if they wish.

Fixes #31.

R: @jeffposnick
CC: @localnerve

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,9 +1,7 @@
-build/
-lib/
 tests/
 node_modules/
 npm-debug.log
 gulpfile.js
 bower.json
-.jshintrc
-.npmignore
+.eslintrc
+
--- a/package.json
+++ b/package.json
@@ -5,7 +5,12 @@
   "scripts": {
     "test": "gulp lint"
   },
+  "main": "lib/sw-toolbox.js",
   "repository": "https://github.com/GoogleChrome/sw-toolbox",
+  "dependencies": {
+    "serviceworker-cache-polyfill": "coonsta/cache-polyfill",
+    "path-to-regexp": "^1.0.1"
+  },
   "devDependencies": {
     "browserify": "^6.3.2",
     "browserify-header": "^0.9.2",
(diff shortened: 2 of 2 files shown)


jeffposnick:
:+1:

jeffposnick:
We'll have to republish eventually—want to cut a new `npm` release now with these changes and a patch version bump, or wait until the cache expiration stuff lands and bump the minor version?

wibblymat:
@jeffposnick @localnerve I tagged 2.1.0 and pushed to npm. I bumped the minor because the networkTimeoutSeconds change came in since 2.0.4.

On a side note, we should stop bumping the version in `package.json` except when actually tagging a release.

PR #43, opened 2015-09-30
Adds a gh-pages gulp task, and a top-level recipes index

### LINKED ISSUE #41: Publish demo/recipes to gh-pages

jeffposnick:
R: @wibblymat @addyosmani 

I've already used this new task to push to https://googlechrome.github.io/sw-toolbox/recipes/

It might need to be tweaked in the future if, e.g., we decide to move to a Jekyll templated build for our recipes, but I am not sure that the set of recipes will ever grow large enough to warrant that overhead.

Closes #41


sw-toolbox.js
```diff
@@ -15,7 +15,7 @@
...
...quire("./helpers"),strategies=require("./strategies");helpers.debug("Service Worker Toolbox is loading");var flatten=function(e){return e.reduce(function(e,t){return e.concat(t)},[])};self.addEventListener("install",function(e){var t=options.cache.name+"$$$inactive$$$";helpers.debug("install event fired"),helpers.debug("creating cache ["+t+"]"),e.waitUntil(helpers.openCache({cache:{name:t}}).then(function(e){return Promise.all(options.preCacheItems).then(flatten).then(function(t){return helpers.debug("preCache list: "+(t.join(", ")||"(none)")),e.addAll(t)})}))}),self.addEventListener("activate",function(e){helpers.debug("activate event fired");var t=options.cache.name+"$$$inactive$$$";e.waitUntil(helpers.renameCache(t,options.cache.name))}),self.addEventListener("fetch",function(e){var t=router.match(e.request);t?e.respondWith(t(e.request)):router["default"]&&"GET"===e.request.method&&e.respondWith(router["default"](e.request))}),module.exports={networkOnly:strategies.networkOnly,networkFirst:strategies.networkFirst,cacheOnly:strategies.cacheOnly,cacheFirst:strategies.cacheFirst,fastest:strategies.fastest,router:router,options:options,cache:cache,uncache:uncache,precache:precache};
```
jeffposnick: It looks like the current `sw-toolbox.js` in master was missing `"GET"===e.request.method`. I'm assuming that was due to me not having https://github.com/GoogleChrome/sw-toolbox/pull/39 merged when I pushed out https://github.com/GoogleChrome/sw-toolbox/pull/37

So this should result in the proper `sw-toolbox.js` making it to master. I hope this doesn't become a recurring problem...

wibblymat: Bower requires you to publish built files in your repo, otherwise we wouldn't bother with them at all.

...but, actually, Bower works off tags, and you can tag non-master branches. We could use something very much like `gulp-gh-pages` to add a publish step that puts the build artefacts in a builds branch and tags that.

I'll open an issue to discuss.

jeffposnick: Alternatively, if we checked in an unminified version of the built library as well, it would be easier to see what's being changed.

PR #50, opened 2015-10-28
Responding with original Network error response incase of cache miss

adityapunjani:
Currently if there is network error such as a 500 and a cache miss we see the error :

`The FetchEvent for "xyz" resulted in a network error response: an object that was not a Response was passed to respondWith()`

This fix passes through correct error response incase of cache miss.


wibblymat:
Awesome, thanks!

PR #51, opened 2015-10-30
Normalize headings and document structure.

jpmedley:
An equivalent PR is being made in sw-precache so that both have roughly similar section titles and order. A copy edit of the content will be in a later PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,9 +184,9 @@ Patches are encouraged, and may be submitted by forking this project and submitt
 
 Copyright 2015 Google, Inc.
 
-Licensed under the [Apache License, Version 2.0[(https://github.com/GoogleChrome
-/sw-toolbox/blob/master/README.md)] (the "License"); you may not use this file
-except in compliance with the License. You may obtain a copy of the License at
+Licensed under the [Apache License, Version 2.0](LICENSE) (the "License"); 
+you may not use this file except in compliance with the License. You may 
+obtain a copy of the License at
 
    http://www.apache.org/licenses/LICENSE-2.0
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -184,9 +184,9 @@ Patches are encouraged, and may be submitted by forking this project and submitt
 
 Copyright 2015 Google, Inc.
 
-Licensed under the Apache License, Version 2.0 (the "License");
-you may not use this file except in compliance with the License.
-You may obtain a copy of the License at
+Licensed under the [Apache License, Version 2.0[(https://github.com/GoogleChrome
+/sw-toolbox/blob/master/README.md)] (the "License"); you may not use this file
```
jeffposnick: I think this should link to https://github.com/GoogleChrome/sw-toolbox/blob/master/LICENSE, which can be simplified in GitHub Markdown to `[Apache License, Version 2.0](LICENSE)`.

jeffposnick:
@wibblymat should weigh in as well before merging, but I think this looks good other than the LICENSE link issue.

wibblymat:
I agree about the LICENSE link, otherwise LGTM

jpmedley:
@jeffposnick @wibblymat 

I might have jumped to a conclusion. It seedmed like both sw-toolkit and sw-precache should have the same license. As they were they both referred to Apache 2.0. 

What you requested, is this what I should do for both?

jeffposnick:
Yep, the same change needs to be made to the `sw-precache` PR.

jpmedley:
@jeffposnick PTAL

jeffposnick:
:+1:

PR #55, opened 2015-11-10
Copy edit the README.md file.

jpmedley:
As promised, this one is a little easier to digest.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ From your registering page, register your service worker in the normal way. For 
 navigator.serviceWorker.register('my-service-worker.js');
 ```
 
-For even lower friction, especiall if you don't do anything more fancy than registering with a default scope, you can instead include the Service Worker Toolbox companion script in your HTML:
+For even lower friction, especially if you don't do anything more fancy than registering with a default scope, you can instead include the Service Worker Toolbox companion script in your HTML:
 
 ```html
 <script src="/path/to/sw-toolbox/companion.js" data-service-worker="my-service-worker.js"></script>
--- a/README.md
+++ b/README.md
@@ -162,7 +162,7 @@ Create a route that causes requests for URLs matching `urlPattern` to be resolve
 
 - `urlPattern` - an Express style route. See the docs for the [path-to-regexp](https://github.com/pillarjs/path-to-regexp) module for the full syntax
 - `handler` - a request handler, as [described above](#using-request-handlers)
-- `options` - an object containing options for the router. This options object will be passed to the request handler. The `origin` option is specific to the router methods, and can be either an exact string or a Regexp against which the origin of the Request must match for the router to be used.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -22,13 +22,13 @@ From your registering page, register your service worker in the normal way. For
 navigator.serviceWorker.register('my-service-worker.js');
 ```
 
-For even lower friction, if you don't intend to doing anything more fancy than just registering with a default scope, you can instead include the Service Worker Toolbox companion script in your HTML:
+For even lower friction, especiall if you don't do anything more fancy than registering with a default scope, you can instead include the Service Worker Toolbox companion script in your HTML:
```
wibblymat: typo: `especiall`

wibblymat:
You changed route to router in a few places, and I'm not 100% sure of all of the changes.

- A **route** is a URL pattern associated with a handler. It defines the behaviour for a section of the site.
- **Routing** is the process of matching an incoming request with the most appropriate **route**.
- The **router** is the component of SW Toolbox that does the **routing**

Given that description, I think that one or more of the changes are incorrect, but I'm perhaps too close to it. I'll leave a separate commit comment to mark the one that I am most convinced is wrong.

README.md
```diff
@@ -156,27 +161,27 @@ Handle the request by trying to fetch the URL from the network. If the fetch fai
 Create a route that causes requests for URLs matching `urlPattern` to be resolved by calling `handler`. Matches requests using the GET, POST, PUT, DELETE or HEAD HTTP methods respectively.
 
 - `urlPattern` - an Express style route. See the docs for the [path-to-regexp](https://github.com/pillarjs/path-to-regexp) module for the full syntax
-- `handler` - a request handler, as [described above](#request-handlers)
-- `options` - an object containing options for the route. This options object will be available to the request handler. The `origin` option is specific to the route methods, and is an exact string or a Regexp against which the origin of the Request must match for the route to be used.
+- `handler` - a request handler, as [described above](#using-request-handlers)
+- `options` - an object containing options for the router. This options object will be passed to the request handler. The `origin` option is specific to the router methods, and can be either an exact string or a Regexp against which the origin of the Request must match for the router to be used.
```
wibblymat: I think the final change on this line should be reverted - it is the route that will or will not be used. The router is always used.

wibblymat:
Otherwise LGTM :+1:

jpmedley:
I reviewed all of them and Fixed:

- `options` - an object containing options for the *route*. This options
object will be passed to the request handler. The `origin` option is
specific to the router methods, and can be either an exact string or a
Regexp against which the origin of the Request must match for the *route*
to be used.



Joe Medley | Technical Writer, DevPlat | jmedley@google.com | 816-678-7195

jpmedley:
FYI, Mat's issue is in another PR, but this one is blocking.

PR #60, opened 2015-12-11
Move to the Google ESLint config, and fix the linting errors.

### LINKED ISSUE #58: Switch to google ESLint config
https://github.com/google/eslint-config-google

wibblymat:
R: @jeffposnick @addyosmani 

This was difficult for me! I personally think that the 80 character line limit is unnecessary, but I've stuck to it anyway. It allowed me to indulge myself in shortening a few variable names (which I like), but I've also just had to manually wrap some lines in awkward places.

There are a few rule overrides currently, I'll open a separate issue to track removing them. The big one is requiring JSDoc comments for things. I think that's a good thing to do, but I didn't want to block the rest of the PR on it.

Fixes #58

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,17 +1,12 @@
 {
-  "extends": "xo",
+  "extends": "google",
   "env": {
     "serviceworker": true,
     "worker": true
   },
   "rules": {
-    "default-case": 0,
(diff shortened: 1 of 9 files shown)


jeffposnick:
:+1:

PR #65, opened 2015-12-21
Fix broken unit tests

### LINKED ISSUE #63: Unit tests broken
The current unit tests are broken for two reasons:

https://github.com/GoogleChrome/sw-toolbox/pull/39 made it so that the `default` route only applies to `GET` requests, and there are a number of tests that exercise the `default` route with other methods, starting at https://github.com/GoogleChrome/sw-toolbox/blob/master/tests/tests.js#L68 Those tests should probably just be removed.

Also, the tests wait on `.ready` instead of waiting until the page is controlled by the service worker, which makes the first test in the suite flaky. (See https://github.com/slightlyoff/ServiceWorker/issues/799).

jeffposnick:
R: @wibblymat @gauntface @addyosmani 

Closes #63, which provides more context.

### DIFF
--- a/tests/tests.js
+++ b/tests/tests.js
@@ -2,6 +2,14 @@
 
 navigator.serviceWorker.register('service-worker.js');
 
+var controlledPromise = new Promise(function(resolve) {
+  if (navigator.serviceWorker.controller) {
+    resolve();
+  } else {
+    navigator.serviceWorker.addEventListener('controllerchange', resolve);
+  }
+});
+
 var checkValue = function(url, value, assert, method) {
   var done = assert.async();
   method = method || 'get';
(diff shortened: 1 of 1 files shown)


tests/tests.js
```diff
@@ -2,6 +2,14 @@
 
 navigator.serviceWorker.register('service-worker.js');
 
+var controlledPromise = new Promise(function(resolve) {
```
gauntface: Could you add a comment here.

I understand the reasoning behind this - but its still hazy and others may not be aware of the reasons.

gauntface:
This looks good - comment would just be a nice to have.

wibblymat:
:+1:

PR #67, opened 2016-01-04
RegExp routes

### LINKED ISSUE #38: Rethink the pathname/origin split when defining handlers
I've never taken to the [ExpressJS-style](http://expressjs.com/guide/routing.html) routes used to define custom handlers. It might make sense for situations in which you're normally matching same-origin URLs and you care about assigning portions of the pathname to request handler parameters, but I find it clumsy for, e.g., matching any request made against a given remote origin.

How about we start accepting a `RegExp` object as the first parameter to the `.get()`, etc. registration handlers? We could still support the old syntax as well. When there's a `RegExp` passed in, we use it as a match against the full `event.request.url` to determine whether the handler applies. We could expose any embedded capture groups in the parameters passed in to the handler, too.

That would allow for, e.g., 

```js
toolbox.router.get(/example\.com.+\.json/, function(request) {
  // handle GET request made to example.com that ends in '.json'
});
```

instead of the current

```js
toolbox.router.get('/(.*).json', function(request) {
  // handle GET request made to example.com that ends in '.json'
}, {origin: /example\.com/});
```

jeffposnick:
R: @addyosmani @wibblymat @gauntface 

I'm a fan of using `RegExp` objects for routing, and having them match the full URL. The current alternative it to use ExpressJS-style strings for path matching, and then use an additional `origin` parameter for matching the origin.

This adds in that functionality, while still supporting ExpressJS-style routing. It also adds in a couple of additional tests and modifies the docs.

I attempted to do this without fully rewriting the internal routing registry code in a slightly-hacky way (it populates the routing registry `Map` object with an entry whose key is the value `RegExp`), but I think that should be safe.

Closes #38


README.md
```diff
@@ -63,7 +89,45 @@ toolbox.router.get(':foo/index.html', function(request, values) {
...
+path. This can lead to simpler cross-origin routing vs. ExpressJS-style routes,
+since both the origin and the path are matched simultaneously, without having
+to specify a separate `origin` option.
+
+Note that while ExpressJS-style routes allow you to name path fragment
+parameters that will be passed to your handler (see `values.foo` in the previous
+example), that functionality is not supported while using regular expression
+routes.
 
+Some examples of using ExpressJS-style routing include:
```
wibblymat: Typo: ExpressJS-style -> Regular Expression style

README.md
```diff
@@ -86,7 +150,12 @@ var myHandler = function(request, values, options) {
 ```
 
 - `request` - The [Request](https://fetch.spec.whatwg.org/#request) object that triggered the `fetch` event
-- `values` - Object whose keys are the placeholder names in the URL pattern, with the values being the corresponding part of the request URL. For example, with a URL pattern of `'/images/:size/:name.jpg'` and an actual URL of `'/images/large/unicorns.jpg'`, `values` would be `{size: 'large', name: 'unicorns'}`
+- `values` - When using Express-style routing paths, this will be an object
```
wibblymat: Consistency - Express or ExpressJS. I think the name of the tool is Express.

recipes/cache-expiration-options/service-worker.js
```diff
@@ -8,10 +8,10 @@
   global.toolbox.options.debug = true;
 
   // Set up a handler for HTTP GET requests:
-  // - '/(.*)' means any URL pathname will be matched.
+  // - /\.ytimg\.com\// will match any requests going against ytimg.com.
   // - toolbox.cacheFirst let us to use the predefined cache strategy for those
   //   requests.
-  global.toolbox.router.get('/(.*)', global.toolbox.cacheFirst, {
+  global.toolbox.router.get(/\.ytimg\.com\//, global.toolbox.cacheFirst, {
```
wibblymat: Nit: The regex is actually too broad, and would match `http://example.com/.ytimg.com/`

README.md
jeffposnick: Fixed.

README.md
jeffposnick: Switched everything to "Express".

recipes/cache-expiration-options/service-worker.js
jeffposnick: It would, but... I have control over the HTTP requests being made, and I'm not concerned about that happening.

jeffposnick:
PTAL.

wibblymat:
:+1:

PR #70, opened 2016-01-12
Adding first take on mocha tests

gauntface:
This a reduced and slightly tidier version of the wider-test-coverage.

Breaking up into smaller PR's so it's a bit more manageable to review.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,13 +14,16 @@
   "devDependencies": {
     "browserify": "^12.0.1",
     "browserify-header": "^0.9.2",
+    "chai": "^3.4.1",
     "eslint": "^1.10.3",
     "eslint-config-google": "^0.3.0",
+    "express": "^4.13.3",
     "gulp": "^3.9.0",
     "gulp-eslint": "^1.1.1",
     "gulp-gh-pages": "^0.5.4",
     "jshint-stylish": "^2.1.0",
     "minifyify": "^7.1.0",
+    "mocha": "^2.3.4",
     "qunitjs": "^1.20.0",
     "temp": "^0.8.3",
     "vinyl-source-stream": "^1.1.0"
--- a/test/browser-tests/precache/precache.js
+++ b/test/browser-tests/precache/precache.js
@@ -1,0 +1,175 @@
+/*
+  Copyright 2014 Google Inc. All Rights Reserved.
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+*/
+
+// This is a test and we want descriptions to be useful, if this
+// breaks the max-length, it's ok.
+
+/* eslint-disable max-len, no-lonely-if */
+/* eslint-env browser, mocha */
+/* global testHelper */
+
+'use strict';
+
+describe('Test precache method', () => {
+  let compareCachedAssets = function(assetList, cachedAssets) {
+    return new Promise((resolve, reject) => {
+      var cachedAssetsKeys = Object.keys(cachedAssets);
+      cachedAssetsKeys.should.have.length(assetList.length);
+
+      for (var i = 0; i < assetList.length; i++) {
+        var key = location.origin + assetList[i];
+        if (typeof cachedAssets[key] === 'undefined') {
+          reject(new Error('Cache doesn\'t have a cache item for: ' + key));
+        }
+
+        cachedAssets[key].status.should.equal(200);
+      }
+
+      resolve();
+    });
+  };
+
+  var serviceWorkersFolder = '/test/browser-tests/precache/serviceworkers';
+
+  it('should precache all desired assets in precache-valid', done => {
+    var assetList = [
+      '/test/data/files/text.txt',
+      '/test/data/files/text-1.txt'
+    ];
+    testHelper.activateSW(serviceWorkersFolder + '/simple.js')
+    .then(() => {
+      return testHelper.getAllCachedAssets('precache-valid');
+    })
+    .then(cachedAssets => {
+      return compareCachedAssets(assetList, cachedAssets);
+    })
+    .then(() => done(), done);
+  });
+
+  it('should precache a single item in precache-valid', done => {
+    var assetList = [
+      '/test/data/files/text.txt'
+    ];
+    testHelper.activateSW(serviceWorkersFolder + '/single-item.js')
+    .then(() => {
+      return testHelper.getAllCachedAssets('precache-valid');
+    })
+    .then(cachedAssets => {
+      return compareCachedAssets(assetList, cachedAssets);
+    })
+    .then(() => done(), done);
+  });
+
+  it('should precache all desired assets from promises in precache-valid', done => {
+    var assetList = [
+      '/test/data/files/text.txt',
+      '/test/data/files/text-1.txt'
+    ];
+    testHelper.activateSW(serviceWorkersFolder + '/promises.js')
+    .then(() => {
+      return testHelper.getAllCachedAssets('precache-valid');
+    })
+    .then(cachedAssets => {
+      return compareCachedAssets(assetList, cachedAssets);
+    })
+    .then(() => done(), done);
+  });
+
+  it.skip('should precache all desired assets from arrays of arrays in precache-valid', done => {
+    var assetList = [
+      '/test/data/files/text.txt',
+      '/test/data/files/text-1.txt',
+      '/test/data/files/text-2.txt',
+      '/test/data/files/text-3.txt',
+      '/test/data/files/text-4.txt',
+      '/test/data/files/text-5.txt',
+      '/test/data/files/text-6.txt',
+      '/test/data/files/text-7.txt',
+      '/test/data/files/text-8.txt'
+    ];
+    testHelper.activateSW(serviceWorkersFolder + '/arrays.js')
+    .then(() => {
+      return testHelper.getAllCachedAssets('precache-valid');
+    })
+    .then(cachedAssets => {
+      return compareCachedAssets(assetList, cachedAssets);
+    })
+    .then(() => done(), done);
+  });
+
+  it.skip('should precache all desired assets from arrays of arrays of promisesin precache-valid', done => {
+    var assetList = [
+      '/test/data/files/text.txt',
+      '/test/data/files/text-1.txt',
+      '/test/data/files/text-2.txt',
+      '/test/data/files/text-3.txt',
+      '/test/data/files/text-4.txt',
+      '/test/data/files/text-5.txt',
+      '/test/data/files/text-6.txt',
+      '/test/data/files/text-7.txt',
+      '/test/data/files/text-8.txt'
+    ];
+    testHelper.activateSW(serviceWorkersFolder + '/promise-arrays.js')
+    .then(() => {
+      return testHelper.getAllCachedAssets('precache-valid');
+    })
+    .then(cachedAssets => {
+      return compareCachedAssets(assetList, cachedAssets);
+    })
+    .then(() => done(), done);
+  });
+
+  it.skip('should not precache paths that do no exist', done => {
+    var testId = 'precache-non-existant-files';
+    var validAssetsList = [
+      '/test/data/files/text.txt',
+      '/test/data/files/text-1.txt'
+    ];
+    testHelper.activateSW(serviceWorkersFolder + '/non-existant-files.js')
+    .then(() => {
+      return testHelper.getAllCachedAssets(testId);
+    })
+    .then(cachedAssets => {
+      return compareCachedAssets(validAssetsList, cachedAssets);
+    })
+    .then(() => done(), done);
+  });
+
+  it('should precache all assets from each install step', done => {
+    var toolboxAssetList = [
+      '/test/data/files/text.txt',
+      '/test/data/files/text-1.txt'
+    ];
+    var additionalInstallAssets = [
+      '/test/data/files/text-2.txt',
+      '/test/data/files/text-3.txt'
+    ];
+    testHelper.activateSW(serviceWorkersFolder + '/custom-install.js')
+    .then(() => {
+      return testHelper.getAllCachedAssets('precache-custom-install-toolbox');
+    })
+    .then(cachedAssets => {
+      return compareCachedAssets(toolboxAssetList, cachedAssets);
+    })
+    .then(() => {
+      return testHelper.getAllCachedAssets('precache-custom-install');
+    })
+    .then(cachedAssets => {
+      return compareCachedAssets(additionalInstallAssets, cachedAssets);
+    })
+    .then(() => done(), done);
+  });
+});
--- a/test/browser-tests/precache/precache.js
+++ b/test/browser-tests/precache/precache.js
@@ -1,5 +1,5 @@
 /*
-  Copyright 2014 Google Inc. All Rights Reserved.
+  Copyright 2016 Google Inc. All Rights Reserved.
 
   Licensed under the Apache License, Version 2.0 (the "License");
   you may not use this file except in compliance with the License.
@@ -17,20 +17,23 @@
 // This is a test and we want descriptions to be useful, if this
 // breaks the max-length, it's ok.
 
-/* eslint-disable max-len, no-lonely-if */
+/* eslint-disable max-len */
 /* eslint-env browser, mocha */
 /* global testHelper */
 
 'use strict';
 
 describe('Test precache method', () => {
-  let compareCachedAssets = function(assetList, cachedAssets) {
-    return new Promise((resolve, reject) => {
-      var cachedAssetsKeys = Object.keys(cachedAssets);
-      cachedAssetsKeys.should.have.length(assetList.length);
+  let compareCachedAssets = (assetList, cachedAssets) => {
+    // We make a set to ensure duplicates are removed from the asset list
+    let assetSet = new Set(assetList);
 
-      for (var i = 0; i < assetList.length; i++) {
-        var key = location.origin + assetList[i];
+    return new Promise((resolve, reject) => {
+      let cachedAssetsKeys = Object.keys(cachedAssets);
+      cachedAssetsKeys.should.have.length(assetSet.size);
+
+      for (let assetPath of assetSet) {
+        let key = location.origin + assetPath;
         if (typeof cachedAssets[key] === 'undefined') {
           reject(new Error('Cache doesn\'t have a cache item for: ' + key));
         }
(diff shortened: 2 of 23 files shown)


package.json
```diff
@@ -14,13 +14,16 @@
   "devDependencies": {
```
jeffposnick: Can you add in https://www.npmjs.com/package/chromedriver as a `devDependency` instead of requiring developers to install it out-of-band? I know that you were including it in the Travis configuration, but it could be nice to explicitly list it here.

I *think* that would also help make sure that it gets cached in between Travis runs, but I don't really understand Travis caching, so I might be wrong.

test/browser-tests/precache/precache.js
```diff
@@ -0,0 +1,175 @@
+/*
+  Copyright 2014 Google Inc. All Rights Reserved.
```
jeffposnick: I was about to write 2015, but... 2016. :calendar: :open_mouth:

test/browser-tests/precache/precache.js
```diff
@@ -0,0 +1,175 @@
...
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+*/
+
+// This is a test and we want descriptions to be useful, if this
+// breaks the max-length, it's ok.
+
+/* eslint-disable max-len, no-lonely-if */
```
jeffposnick: Do you need to disable `no-lonely-if`? I don't see where it's an issue in this file, and if it is an issue, it would be better to just stick to the common style.

test/browser-tests/precache/precache.js
```diff
@@ -0,0 +1,175 @@
...
+/* eslint-env browser, mocha */
+/* global testHelper */
+
+'use strict';
+
+describe('Test precache method', () => {
+  let compareCachedAssets = function(assetList, cachedAssets) {
```
jeffposnick: Nit: You're using arrow functions pretty much everywhere else, so it looks out of place to use the `function()` syntax here.

test/browser-tests/precache/precache.js
```diff
@@ -0,0 +1,175 @@
...
+        }
+
+        cachedAssets[key].status.should.equal(200);
+      }
+
+      resolve();
+    });
+  };
+
+  var serviceWorkersFolder = '/test/browser-tests/precache/serviceworkers';
```
jeffposnick: You're using `let` elsewhere in this file, so how about `const` here?

test/browser-tests/precache/precache.js
```diff
@@ -0,0 +1,175 @@
...
+      '/test/data/files/text-1.txt'
+    ];
+    testHelper.activateSW(serviceWorkersFolder + '/simple.js')
+    .then(() => {
+      return testHelper.getAllCachedAssets('precache-valid');
+    })
+    .then(cachedAssets => {
+      return compareCachedAssets(assetList, cachedAssets);
+    })
+    .then(() => done(), done);
```
jeffposnick: `.then(() => done(), done)` can be rewritten as `.then(done, done)` here and elsewhere.

test/browser-tests/precache/precache.js
```diff
@@ -0,0 +1,175 @@
...
+  });
+
+  it.skip('should precache all desired assets from arrays of arrays in precache-valid', done => {
```
jeffposnick: I see this test and a few others marked with `skip()`. Are they ready for review? If they get checked in without review, they might just fall through the cracks.

If they need to be checked in as part of this PR but aren't being run, could you elaborate with a comment?

test/browser-tests/precache/serviceworkers/arrays.js
```diff
@@ -0,0 +1,51 @@
...
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+*/
+
+'use strict';
+
+/* eslint-env browser */
```
jeffposnick: `/* eslint-env browser */` → `/* eslint-env worker, serviceworker */`

test/browser-tests/precache/serviceworkers/arrays.js
```diff
@@ -0,0 +1,51 @@
...
+self.toolbox.precache([
+  '/test/data/files/text.txt',
+  [
+    '/test/data/files/text-1.txt',
+    '/test/data/files/text-2.txt'
+  ],
+  [
+    '/test/data/files/text-3.txt',
+    [
+      '/test/data/files/text-4.txt',
```
jeffposnick: Going all the way down to, it looks like, 6 levels of nesting doesn't seem necessary to properly test the code. Would 2-3 levels of nesting be sufficient to cover the edge cases?

test/browser-tests/precache/precache.js
```diff
@@ -0,0 +1,175 @@
...
+/* eslint-env browser, mocha */
+/* global testHelper */
+
+'use strict';
+
+describe('Test precache method', () => {
+  let compareCachedAssets = function(assetList, cachedAssets) {
+    return new Promise((resolve, reject) => {
+      var cachedAssetsKeys = Object.keys(cachedAssets);
+      cachedAssetsKeys.should.have.length(assetList.length);
```
jeffposnick: If there are repeated entries in `assetList`, then it's possible that `assetList` and `cachedAssetKeys` will have the same length despite there being more unique entries in `cachedAssetKeys`.

If you're don't care about testing scenarios in which `cachedAssetKeys` has additional entries that aren't accounted for in `assetList`, then you should remove this assertion.

If you do care about testing that `assetList` and `cachedAssetKeys` are exactly the same, then you should remove duplicates from `assetList` before asserting on the lengths. Using a `Set` might help.

test/browser-tests/precache/serviceworkers/promise-arrays.js
```diff
@@ -0,0 +1,51 @@
...
+self.toolbox.precache([
+  Promise.resolve('/test/data/files/text.txt'),
+  [
+    Promise.resolve('/test/data/files/text-1.txt'),
+    Promise.resolve('/test/data/files/text-2.txt')
+  ],
+  [
+    Promise.resolve('/test/data/files/text-3.txt'),
+    [
+      Promise.resolve('/test/data/files/text-4.txt'),
```
jeffposnick: Similar comment re: required nesting depth as in the other file.

test/browser-tests/precache/serviceworkers/promises.js
```diff
@@ -0,0 +1,31 @@
...
+
+/* eslint-env browser */
+
+importScripts('/sw-toolbox.js');
+
+self.toolbox.options.cache = {
+  name: 'precache-valid'
+};
+
+// Imagine getting a list of assets from an API to cache
```
jeffposnick: This comment sounds like a non-sequitur.

test/browser-tests/precache/serviceworkers/simple.js
```diff
@@ -0,0 +1,30 @@
...
+
+/* eslint-env browser */
+
+importScripts('/sw-toolbox.js');
+
+self.toolbox.options.cache = {
+  name: 'precache-valid'
+};
+
+self.toolbox.precache([
```
jeffposnick: I don't see any tests of a mix of `Promise`s and `string`s in the same array. An additional test for that mix seems more likely to find bugs and edge cases.

test/index.html
```diff
@@ -0,0 +1,129 @@
...
+      // bypassing the tests altogether.
+      // This is desirable to allow travis to run tests in all browsers
+      // regardless of support or not and perform tests when the browser
+      // starts to support service workers.
+      if (!('serviceWorker' in navigator)) {
+        publishTestResults();
+        return;
+      }
+
+      window.chai.should();
```
jeffposnick: Can you add a comment explaining why this is needed?

test/index.html
```diff
@@ -0,0 +1,129 @@
...
+
+      // We unregister all service workers, clear all caches and remove
+      // All registered iframes
+      beforeEach(function(done) {
```
jeffposnick: See comment elsewhere about using https://mochajs.org/#working-with-promises instead of explicitly adding `.then(done, done)` to your promise chains.

test/index.html
```diff
@@ -0,0 +1,129 @@
...
+      after(function(done) {
+        testHelper.cleanState()
+        .then(() => done(), done);
+      });
+
+      var runResults = mocha.run();
+
+      // If there are no tests we should define window.swtoolbox
+      // so the automated test suite will know the test has ended
+      if (runResults.total === 0) {
```
jeffposnick: Why would this happen? It sounds like an error condition, so wouldn't it be better to report some sort of failure, either explicitly or implicitly via the timeout?

test/libs/helper-functions.js
```diff
@@ -0,0 +1,194 @@
...
+var getIframe = function() {
+  return new Promise(resolve => {
+    var existingIframe = document.querySelector('.js-test-iframe');
+    if (existingIframe) {
+      return resolve(existingIframe);
+    }
+
+    var newIframe = document.createElement('iframe');
+    newIframe.classList.add('js-test-iframe');
+    newIframe.src = '/test/iframe/' + Math.random();
```
jeffposnick: Could you use a globally-scoped `iframeId` counter that gets incremented after each iframe is created? If you don't think that's sufficient because of the caching concerns in between runs, then combine the counter with `Date.now()`.

Including `Math.random()` in a test introduces an infinitesimal chance of unnecessary flakiness.

test/libs/helper-functions.js
```diff
@@ -0,0 +1,194 @@
...
+    });
+    document.body.appendChild(newIframe);
+  });
+};
+
+window.testHelper = {
+  unregisterAllRegistrations: function() {
+    return navigator.serviceWorker.getRegistrations()
+      .then(registrations => {
+        if (registrations.length === 0) {
```
jeffposnick: This function body could be simplified to

```js
return Promise.all(registrations.map(registration => {
  return registration.unregister();
}));
```

test/libs/helper-functions.js
```diff
@@ -0,0 +1,194 @@
...
+          );
+        }
+        return Promise.all(unregisterPromises);
+      });
+  },
+
+  clearAllCaches: function() {
+    return window.caches.keys()
+      .then(cacheNames => {
+        if (cacheNames.length === 0) {
```
jeffposnick: This function body could be simplified to

```js
return Promise.all(cacheNames.map(window.caches.delete));
```

Or if you'd rather be explicit about the parameter passed to `caches.delete()`, you can use

```js
return Promise.all(cacheNames.map(cacheName => window.caches.delete(cacheName)));
```

test/libs/helper-functions.js
```diff
@@ -0,0 +1,194 @@
...
+        // always trigger an install state change
+        registration.installing.onstatechange = function() {
+          if (this.state !== 'installed') {
+            return;
+          }
+
+          resolve(iframe);
+        };
+      })
+      .catch(err => {
```
jeffposnick: I don't think this is needed. An uncaught exception inside your promise's function will implicitly trigger a rejection, without you having to do anything.

test/libs/helper-functions.js
```diff
@@ -0,0 +1,194 @@
...
+        // self.clients.claim()
+        registration.installing.onstatechange = function() {
+          if (this.state !== 'activated') {
+            return;
+          }
+
+          resolve(iframe);
+        };
+      })
+      .catch(err => {
```
jeffposnick: See previous comment about not needing this.

test/libs/helper-functions.js
```diff
@@ -0,0 +1,194 @@
...
+      });
+    });
+  },
+
+  // This is a helper method that checks the cache exists before
+  // getting all the cached responses.
+  // This is limited to text at the moment.
+  getAllCachedAssets: function(cacheName) {
+    var cache = null;
+    return window.caches.keys()
```
jeffposnick: You can use `window.caches.has(cacheName)`: https://developer.mozilla.org/en-US/docs/Web/API/CacheStorage/has

package.json
gauntface: I tried using chrome driver with install locally and it didn't work, it had to be installed globally. Can confirm when I come back round to setting up travis (Will be the next PR).

test/libs/helper-functions.js
```diff
@@ -0,0 +1,194 @@
...
+      .catch(err => {
+        reject(err);
+      });
+    });
+  },
+
+  // This is a helper method that checks the cache exists before
+  // getting all the cached responses.
+  // This is limited to text at the moment.
+  getAllCachedAssets: function(cacheName) {
```
jeffposnick: `getAllCachedAssets()` resolves to an `Object` with keys that consist of URLs, and values that consist of `Response`s. As far as I can tell, though, you never make use of the `Response` values anywhere, and just use `Object.keys()` to extract out the URLs from the resolved value.

Can you just use an array of URLs as the resolution value? You can get those pretty easily via `cache.keys()` and save yourself a lot of trouble.

test/server/index.js
```diff
@@ -0,0 +1,57 @@
...
+// header to any service worker loaded in the tests. This allows the scope
+// to be manipulated any way we want / need during testing.
+
+var path = require('path');
+var express = require('express');
+var app = express();
+
+// Set up static assets
+app.use('/test/browser-tests/',
+  express.static(path.join(__dirname, '../browser-tests/'), {
```
jeffposnick: Here and elsewhere: when using `path.join()`, pass in all of the path components one by one as separate arguments. Passing in `'../browser-tests/'` defeats the benefit of using `path.join()`, since it assumes that `/` is the path delimiter.

So this should be `path.join(__dirname, '..', 'browser-tests')`

test/browser-tests/precache/precache.js
gauntface: If the promise resolves with a value the done will receive that parameter and that results in the test failing.

test/server/index.js
```diff
@@ -0,0 +1,57 @@
...
+
+// This server is needed most importantly to add the Service-Worker-Allowed
+// header to any service worker loaded in the tests. This allows the scope
+// to be manipulated any way we want / need during testing.
+
+var path = require('path');
+var express = require('express');
+var app = express();
+
+// Set up static assets
```
jeffposnick: Would it make sense to include headers to disable browser caching for everything that gets served?

package.json
jeffposnick: It should work if you start the test with `npm test`. It won't work if you start the test via `gulp`.

jeffposnick:
Reviewed.

test/browser-tests/precache/precache.js
gauntface: I checked them in like this since they are fail and likely require changes to the code to either throw errors or a change in the API surface to not allow them (in which case the test changes to check for the graceful failure).

I'll remove the skip but it does means tests will fail.

test/browser-tests/precache/serviceworkers/arrays.js
gauntface: When I tested it I think it went to three levels (i.e. the test cached 3 levels deep).

I see no reason not to stress test like this when it's so cheap to do so.

test/browser-tests/precache/serviceworkers/simple.js
gauntface: For the most part these tests are copying the qunit tests from what I could see, would you want me to add that test in for this PR or a seperate PR?

test/index.html
gauntface: runResults.total (i.e. the number of tests to be run) is set my mocha - but the results of pass and fail can't be known until the tests have actually been run.

The ONLY reason for the publishTestResults is to ensure that the selenium tests end as soon as possible. This was likely added to account for no tests in browsers that don't support service worker. This is still a handy fail safe, but agreed, serves little purpose for our tests. Will remove.

test/libs/helper-functions.js
gauntface: My and Paul Lewis have found scenarios where nested callbacks or promises can result in the error getting swallowed and the promise not resolving.

test/libs/helper-functions.js
gauntface: Response.status is used: https://github.com/GoogleChrome/sw-toolbox/blob/6a6194c0d63e86e207b95cd72e748f1ecfc59104/test/browser-tests/precache/precache.js#L38

Tests could also be extending in the future to test the size of the response to see if it's correct (Or compare text response).

gauntface:
Thanks for the awesome feedback @jeffposnick - gone through and addressed the majority of it I think.

test/libs/helper-functions.js
jeffposnick: Was that behavior an expected edge case, or a bug? Can you reproduce it?

Alternatively, you can just return a promise chain that starts with `getIframe()` instead of constructing a new `Promise` wrapper.

test/browser-tests/precache/serviceworkers/simple.js
jeffposnick: Separate PR is fine.

test/libs/helper-functions.js
gauntface: It was reproducible when we found it. Don't think we documented it
anywhere.

test/browser-tests/precache/serviceworkers/arrays.js
jeffposnick: I'd recommend using the minimum amount of test data to properly exercise each code path.

test/libs/helper-functions.js
jeffposnick: Okay, fair enough.

test/browser-tests/precache/serviceworkers/arrays.js
gauntface: To me properly exercising == stress testing / worst case scenario.

test/libs/helper-functions.js
gauntface: I can do that - presumably I'd have to wrap this line in a promise as a result? Still preferred?

https://github.com/GoogleChrome/sw-toolbox/blob/mocha-tests/test/libs/helper-functions.js#L93

(Super vague example of why I just use reject from this point: http://jsbin.com/nuyiwoxaxa/edit?html,js,console)

test/server/index.js
gauntface: I've made the change, but could you explain to me why it defeats the point?

I use it just to safely handle directory slashes and fix relative paths, sounds like I missing some kind of behaviour?

gauntface: Also would that result in:

    res.sendFile(path.join(__dirname, '../data/test-iframe.html'));

becoming:

      res.sendFile(path.join(__dirname, '..', 'data', 'test-iframe.html'));

test/server/index.js
gauntface: On the fence with this.

For now, the tests built in such a way that we don't need to account for caching scenarios (i.e. at no stage are the assets be altered, they are always the same, so caching isn't an issue).

If we get to a stage of testing service upgrade paths then caching manipulation becomes important and in that case we'll likely want ways of enabling / disabling caching to ensure the behavior is consistent.

Happy to add it - but I don't feel like it's useful at this stage.

gauntface:
@jeffposnick I think I addressed the majority of your comments.

test/libs/helper-functions.js
jeffposnick: The additional effort of wrapping a promise around the `onstatechange` handler would probably end up being messier :grimacing: 

While it's been a while since I've read through [@jakearchibald's article, in which you're featured prominently](https://jakearchibald.com/2015/tasks-microtasks-queues-and-schedules/), I think I understand why Promise 1's `.catch()` isn't triggered in your JSBin, and that doesn't *seem* like a bug.

But I think my original point still applies: in scenarios in which the `.catch()` **is** triggered, explicitly calling `reject(err)` inside the `.catch()` will just match the implicit behavior that you'd get without a `.catch()`.

test/server/index.js
jeffposnick: Okay, feel free to omit.

jeffposnick:
:+1: with the caveat that I'm still not sold on the necessity of explicitly calling `reject()`. But it's not a big deal.

test/server/index.js
jeffposnick: What I meant was that `path.join()` will construct paths using `path.sep` as the delimiter, which is [normally what you'd want](https://gist.github.com/domenic/2790533#paths-and-urls) to ensure that your code works on both Windows and UNIX-y platforms.

But if you pass in some strings that have `/` hardcoded in them already, you lose that benefit, since you'll end up with a string containing a mix of `\` and `/` on Windows. That's why I'm suggesting that you pass in each path component as a separate argument.

gauntface: Aaahhhh clever. Got it. Thank you. I got all those things fixed anyway.

wibblymat:
I checked out the branch, but the tests don't actually pass. Is that just because of the ones that where using `.skip()` are now aren't?

I think that having a skipped test is fine, providing:
- An issue is open to fix what ever problem causes the test to fail
- A comment is added to the test linking back to the issue

On the other hand, if the tests are all supposed to pass then something is wrong :)

gauntface:
Apologies - I hit merge before your comment came up (github went a bit weird).

It's a chicken and egg situation regarding breaking tests - but will raise bugs for each failing test.

PR #78, opened 2016-01-14
Test for precache mixture of promises, strings and arrays

gauntface:
cc @jeffposnick @wibblymat 

As discussed a test for the mix of strings, promises and arrays in precache method.


jeffposnick:
:+1:

Issue #77, opened 2016-01-14
Proposal for precache API simplification


gauntface:
cc @addyosmani @wibblymat @jeffposnick 

At the moment the precache API allows strings, arrays and promises.

I would like to propose the following:

Change precache to accept only an array, this will cause:

    toolbox.precache('/single-asset);

to be written as:

    toolbox.precache(['/single-asset']);

Secondly remove support for nested arrays. This would require the developer to flatten the arrays and means the API wouldn't need to have some arbitrary limit on the number of nests.

     toolbox.precache(['/single-asset', ['/asset-nested', '/asset-nested-2']]);

becomes:

         toolbox.precache(['/single-asset', '/asset-nested', '/asset-nested-2']);

I don't feel like asking dev to do this is too much of an ask (lodash has a deepflatten method if consumers of the library wanted a quick solution).

I feel like the majority of users will want several assets cached rather than a single asset for the majority of the time.

I'm struggling what a good reasoning for promises in precache. I can see some value in being able to get an asset list, are there other use cases?

The main reason I see it as being beneficial is largely because the precache method abstracts away the ability to use event.waitUntil , so this support is your only option to perform some async action.

Ideally I'd like to see either precache take a single promise i this scenario and expect consumers of the library to make user of Promise.all or whatever means to manage logic required OR make it such that toolbox caching can be done from within the install step. 

For the single promise approach the expectation of this would be to return an array of strings to cache. This would ideally be named something differently to avoid overloading the precache method.

This would result in an API surface of something along the lines of:

toolbox.precache(<Array of Strings>);
toolboxprecacheAsync(<Promise which resolves to an array of strings>);

jeffposnick:
Some background (not much context, but still) from the old repo: https://github.com/wibblymat/shed/issues/21

`<platinum-sw>` does make use of passing in a `Promise` to support the asset list use case: https://github.com/PolymerElements/platinum-sw/blob/master/bootstrap/sw-toolbox-setup.js#L25

I'm honestly not the biggest fan of the `toolbox.precache()` name, since it causes confusion with regards to `sw-precache`. I wouldn't object to a broader renaming.

gauntface:
Happy for a broader renaming, not sure what a good alternative is though - precache describe the action of the method.

Regarding the seperation of calls, for or against @jeffposnick @wibblymat @addyosmani @jakearchibald?

toolbox.precache(<Array of Strings>);
toolboxprecacheAsync(<Single Promise>);

[...]

PR #81, opened 2016-01-15
Automated Tests with WebDriver + Travis Integration

gauntface:
@wibblymat @jeffposnick @addyosmani 

The major concern I have with these tests is that to configure Chrome Stable, Chrome Beta, Firefox Stable and Firefox beta, I've had to define the path of the binary for nearly all of them.

This is tying the tests to linux at the moment, which is what Travis is using. I'm happy to explore trying to get this to work for OS X and Windows if desired, but in all honestly, the automated tests being run locally aren't too useful, the windows pop up and disappear before you can do anything useful.


addyosmani:
It's quite useful to be able to validate the tests in other channels of our target browsers. Thanks for getting this setup for Travis, @gauntface. 

Speaking of - Travis appears to be failing the same tests in Chrome beta and stable atm.

gauntface:
Tests are failing because the expected behaviour isn't happening (hence proposal for API change: https://github.com/GoogleChrome/sw-toolbox/issues/77).

I had already made them skips so that the tests would pass until the appropriate behaviour was determined but that also hides the problem (chicken and egg situation).

.travis.yml
```diff
@@ -0,0 +1,24 @@
+# You must sign into travis-ci.org and set the commit hook on your project for travis to
```
jeffposnick: Who's the audience for this comment? Us, as the owners of this repo? Do we need to add `secure:` somewhere?

I don't have any experience doing this sort of thing, but there are probably other folks who will end up being similarly confused.

If there's an existing web page in the Travis docs that provides more context and explicit instructions, could you link to it?

.travis.yml
```diff
@@ -0,0 +1,24 @@
...
+# run on your project. The secure: variable must be generated by running `travis encrypt`
+# on a github oauth key that you can generate using curl.
+
+sudo: false
+language: node_js
+node_js:
+  - 'stable'
+addons:
+  firefox: latest
```
jeffposnick: It [looks like](https://docs.travis-ci.com/user/firefox/#Version-aliases) there's a `latest-beta` alias that corresponds to the Firefox Beta release. Does `.travis.yml` let you supply a list of aliases here, or are you forced to choose only one of `latest` and `latest-beta`?

test/automated-suite.js
```diff
@@ -0,0 +1,217 @@
...
+    // The driver methods are wrapped in a new promise because the
+    // selenium-webdriver API seems to using some custom promise
+    // implementation that has slight behaviour differences.
+    return new Promise((resolve, reject) => {
+      driver.get('http://localhost:8888/test/')
+      .then(() => {
+        return driver.executeScript('return window.navigator.userAgent;');
+      })
+      .then(userAgent => {
+        // This is just to help with debugging so we can get the browser version
```
jeffposnick: :+1:

test/automated-suite.js
```diff
@@ -0,0 +1,217 @@
...
+        console.log('    Browser User Agent [' + browserName + ']: ' + userAgent);
+      })
+      .then(() => {
+        // We get webdriver to wait until window.swtoolbox.testResults is defined.
+        // This is set in the in browser mocha tests when the tests have finished
+        // successfully
+        return driver.wait(function() {
```
jeffposnick: Putting in an explicit timeout set to whatever you think is reasonable would be clearer than relying on whatever the default timeout value is.

test/automated-suite.js
```diff
@@ -0,0 +1,217 @@
...
+      .thenCatch(function(err) {
+        reject(err);
+      });
+    })
+    .then(testResults => {
+      if (testResults.failed.length > 0) {
+        var failedTests = testResults.failed;
+        var errorMsg = 'Issues in ' + browserName + '.\n\n' + browserName + ' had ' + testResults.failed.length + ' test failures.\n';
+        errorMsg += '------------------------------------------------\n';
+        for (var i = 0; i < failedTests.length; i++) {
```
jeffposnick: As long as we're assuming ES2015, how about:

```js
errorMessage += failedTests.map((failedTest, i) => {
  return `[Failed Test ${i}]\n    ${testResult.title}\n`;
}).join('\n');
```

jeffposnick:
Reviewed. Thanks for your continued work to get things automated!

.travis.yml
gauntface: When I tried adding both, it seemed like the firefox beta install would override the firefox stable install (most likely because it was the last one defined).

Looking online everyone suggests that to have both at the same time on a linux machine, you need to save the firefox beta tar somewhere different to the stable rather than get it through a package manager.

jeffposnick: Cool. Just checking.

test/automated-suite.js
gauntface: This isn't a timeout. Mocha will have a global timeout configured in the gulp plugin that will end a test when reached. This is to wait until an event has occured inside the browser.

In this case it waits for a javascript variable to be assigned as a signal to know that the tests have finished. I couldn't find any other reference for detecting when an event has occurred within the browser. Happy to change it for something else if anyone knows of a better way.

.travis.yml
gauntface: Likely just something I picked up from travis docs originally.

gauntface:
Ta for feedback, just addressed or responded to current PR comments.

test/automated-suite.js
jeffposnick: Gotcha. Relying on the timeout from the `gulp` plugin sounds good.

jeffposnick:
:+1:

test/automated-suite.js
```diff
@@ -0,0 +1,212 @@
...
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+*/
+
+'use strict';
+
+// This is a test and we want descriptions to be useful, if this
```
wibblymat: Nit - this comment has become separated from the ESLint directives that it refers to

test/automated-suite.js
```diff
@@ -0,0 +1,212 @@
...
+          return `[Failed Test ${i + 1}]\n    ${failedTest.title}\n`;
+        }).join('\n');
+        errorMessage += '------------------------------------------------\n';
+        throw new Error(errorMessage);
+      }
+    });
+  };
+
+  it('should pass all tests in Chrome Stable', done => {
+    checkFileExists('/usr/bin/google-chrome-stable')
```
wibblymat: FYI, you can use the npm module `which` to get the path of a binary that is in the `PATH`, and therefore check that it exists. Assuming that everything gets installed in the `PATH` on Travis this should make things work a little bit more cross platform.

Fine as a follow-up though, as getting things working on Travis is the goal here.

test/automated-suite.js
```diff
@@ -0,0 +1,212 @@
...
+      })
+      .then(() => {
+        // This simply retrieves the test results from the inbrowser mocha tests
+        return driver.executeScript('return window.swtoolbox.testResults;');
+      })
+      .then(testResults => {
+        // Resolve the outer promise to get out of the webdriver promise chain
+        resolve(testResults);
+      })
+      .thenCatch(function(err) {
```
wibblymat: `thenCatch`? Is that a typo?

Seems this could be just `.catch(reject)`

wibblymat:
Awesome!

test/automated-suite.js
gauntface: Nah this is webdrivers weird promises that it uses. I wrap it in a promise to contain this as much as possible.

gauntface:
Ta all for feedback. Should all be addressed now.

gauntface:
Updated with master and addressed some offline comments from @wibblymat

wibblymat:
I guess the last piece of the puzzle is getting the tests to pass. My feeling is that the tests in this PR should change to reflect the current behaviour, and updated tests should be included in whatever PR fixes #77.

WDYT? @jeffposnick @gauntface

jeffposnick:
I'm fine with merging this PR as-is.

wibblymat:
:dancers: :shipit:

PR #84, opened 2016-01-20
Reject networkFirst when there's a network failure + cache miss

### LINKED ISSUE #80: Can I serve an Offline page?
I currently have

```
toolbox.precache([
    '/Offline',
    '/Offline/Json'
]);

toolbox.router.get('/(.*)', toolbox.networkFirst);
```

How easy is it for me to serve the offline page if the requested HTML is not in the cache already?

jeffposnick:
R: @gauntface @wibblymat @addyosmani 

This closes #80 by ensuring that `networkFirst` rejects when the following conditions are met:
- There's no previously cached entry to return.
- The initial network request rejected (as opposed to resolved with an error `Response`).

Prior to this PR, `networkFirst` would resolve with `undefined` when those conditions were met. This PR bring the `networkFirst` behavior in line with the default `fetch()` behavior.

I'm very much tempted to add a test case for this scenario, but I don't know how to properly force a `NetworkError` in response to a `fetch()`. Maybe once there's some precedent for mocking in our tests that could be revisited.


gauntface:
There are tests along those lines in the branch the original tests were
written to, they just haven't made it to master yet. Happy to add a test
case separately from this PR.

addyosmani:
> This closes #80 by ensuring that networkFirst rejects when the following conditions are met: There's no previously cached entry to return. The initial network request rejected (as opposed to resolved with an error Response).

The updated behaviour LGTM. Requires rebase.

wibblymat:
I held off LGTMing this for ages because of the old "is it really an exception case?" dilemma. But 5 days is long enough to convince me that I don't have a strong enough opinion to challenge it.

LGTM!

wibblymat:
@googlebot I don't know what confused you, but no one involved here needs to sign a CLA anyway :)

wibblymat:
@googlebot confirm

Or whatever the magic incantation needs to be

AaronLayton:
If I update in Bower will I get this update now?

jeffposnick:
@AaronLayton, feel free to pull in `GoogleChrome/sw-toolbox#master` in your `bower.json` to test out the updated behavior. We haven't tagged a formal release for this yet.

@wibblymat, I think the fact that the Fetch API treats this type of failure as an exception means that it's a little more clear-cut as to what `sw-toolbox` should do. We're not translating an error `Response` to an exception here; we're just passing along an exception that was triggered by the underlying `fetch()`.

PR #88, opened 2016-02-02
First set of router.get tests

gauntface:
cc @wibblymat @jeffposnick

### DIFF
--- a/test/browser-tests/router-get/router-get.js
+++ b/test/browser-tests/router-get/router-get.js
@@ -120,6 +120,16 @@ describe('Test router.get method', () => {
     .catch(done);
   });
 
+  // Firefox version 45+ support fetch requests to other origins going through
+  // service workers. This check will skip tests for older version of
+  // firefox and reenable the tests when appropriate.
+  var firefoxVersion = /Firefox\/(\d+).\d+/.exec(navigator.userAgent);
+  if (firefoxVersion && parseInt(firefoxVersion[1], 10) < 45) {
+    console.warn('Tests skipped due to version of Firefox not supporting ' +
+      'cross origin requests via fetch()');
+    return;
+  }
+
   it('should not match relative path starting with the origin defined in toolbox route. Origin option as regex.', done => {
     performTest(
       serviceWorkersFolder + '/origin-matching.js',
--- a/test/browser-tests/router-get/router-get.js
+++ b/test/browser-tests/router-get/router-get.js
@@ -14,18 +14,17 @@
   limitations under the License.
 */
 
-// This is a test and we want descriptions to be useful, if this
-// breaks the max-length, it's ok.
-
-/* eslint-disable max-len */
 /* eslint-env browser, mocha */
-/* global testHelper */
 
 'use strict';
 
 describe('Test router.get method', () => {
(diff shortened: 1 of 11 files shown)


test/browser-tests/router-get/router-get.js
```diff
@@ -0,0 +1,296 @@
...
+  See the License for the specific language governing permissions and
+  limitations under the License.
+*/
+
+// This is a test and we want descriptions to be useful, if this
+// breaks the max-length, it's ok.
+
+/* eslint-disable max-len */
+/* eslint-env browser, mocha */
+/* global testHelper */
```
jeffposnick: Optional: If you think you'll be using `testHelper` throughout the tests under `/browser-tests/`, you could move this config into a separate directory-scoped `.eslintrc` file. (E.g. [this](https://github.com/GoogleChrome/samples/blob/gh-pages/service-worker/.eslintrc) from the Chrome samples project.)

test/browser-tests/router-get/router-get.js
```diff
@@ -0,0 +1,296 @@
...
+
+      // Call the iframes fetch event so it goes through the service worker
+      return performFetch(newIframe, fetchUrl, expectedString);
+    });
+
+    if (done) {
+      testPromise = testPromise.then(() => done(), done);
+    }
+
+    testPromise = testPromise.then(() => iframe);
```
jeffposnick: I don't understand what this accomplishes. Could you add a comment clarifying why this is needed?

test/browser-tests/router-get/serviceworkers/definition-order.js
```diff
@@ -0,0 +1,37 @@
...
+  limitations under the License.
+*/
+
+'use strict';
+
+/* eslint-env worker, serviceworker */
+
+importScripts('/sw-toolbox.js');
+
+self.addEventListener('install', function() {
```
jeffposnick: Optional: since you're doing the same `skipWaiting()` + `clients.claim()` setup in each SW file, you can abstract that code out into a separate `sw-setup.js` file, and then add it to the array passed to `importScripts()`.

test/browser-tests/router-get/router-get.js
```diff
@@ -0,0 +1,296 @@
...
+    performTest(
+      serviceWorkersFolder + '/full-url.js',
+      location.origin + '/test/absolute-url-test',
+      '/test/absolute-url-test',
+      done
+    );
+  });
+
+  // TODO: Find out correct behaviour https://github.com/GoogleChrome/sw-toolbox/issues/86
+  it.skip('should throw an error for route with an origin defined in sw', done => {
```
jeffposnick: I know that this is a skipped test, but this has the same description as the previous test. Could you change it to distinguish how this test is different?

jeffposnick:
Reviewed. Thanks!

gauntface:
Thanks @jeffposnick gone through and made fixes / changes on your suggestions.

Completely right on the eslint and skip and claim comments - much nicer.

wibblymat:
@gauntface it looks like you didn't add the new `.eslintrc` file that was mentioned in a comment. Other than that, this LGTM.

jeffposnick:
:+1:

Issue #79, opened 2016-01-14
toolbox.precache Array of Arrays of Promise causes incorrect caching of URL


gauntface:
If you nest promises, no error is thrown and precache will attempt to cache the following:

http://localhost:8888/test/browser-tests/precache/serviceworkers/[object%20Object]

See the precache/servoceworkers/promise-array.js test

Issue #74, opened 2016-01-14
toolbox.precache Arrays of Arrays Need Removing of error on X levels deep


gauntface:
At the moment you can do the following:

    self.toolbox.precache([
      '/test/data/files/text.txt',
      [
        '/test/data/files/text-1.txt',
        '/test/data/files/text-2.txt'
      ],
      [
        '/test/data/files/text-3.txt',
        [
          '/test/data/files/text-4.txt',
          [
            '/test/data/files/text-5.txt',
            [
              '/test/data/files/text-6.txt',
              [
                '/test/data/files/text-7.txt',
                [
                  '/test/data/files/text-8.txt'
                ]
              ]
            ]
          ]
        ]
      ]
    ]);

Based on the test results, this will result in 5 assets being cached, presumably text.txt -> text-4.txt. Either this API should be simplified to support a single array level of assets and error if it finds nested arrays or it needs recursively loop through nested arrays.

Personally - I would say it should only include support for a single array.

PR #90, opened 2016-02-16
Precache checks for input and validates the resolved input + tests

gauntface:
@wibblymat @jeffposnick 

Should address:
https://github.com/GoogleChrome/sw-toolbox/issues/74
https://github.com/GoogleChrome/sw-toolbox/issues/79
https://github.com/GoogleChrome/sw-toolbox/issues/77

At the moment the API is simply precache for an Array or Promise. Happy to seperate to precache() and precacheAsync() depending on what you think is best.


wibblymat:
I'm not confident these changes are actually all good.

I know it's discussed in the other issues, but my understanding is that we need to support
- An Array of strings
- A Promise for an Array of strings

all other combinations are superfluous. For example, if the developer has an array of Promises they can themselves pass `Promise.all(arr)`.

The behaviour I'm seeing in the PR means that you cannot have a Promise for an Array, which is an important feature. For example, if you want to get the list of assets from IDB.

This should then be pretty easy to implement
```javascript
function precache(resources) {
  if ((!(resources instanceof Promise) && !Array.isArray(resources)) {
    resources = [resources]
  }
  precacheItems.push(resources)
}

...

cache.addAll(
  Promise.all(precacheItems)
    .then(items => [].concat(...items))
).catch(e => 'Cache.addAll failed for one of the resources')
```

Also, it would be nice if we accepted strings *or* `Request` objects, like `cache.addAll` itself does.

This let's you set options for how the resource should be fetched, like whether credentials should be passed, or whether to make a CORS request.

```javascript
var req = new Request('https://other.origin/image.png', {mode: 'cors'});
toolbox.precache(request);
```

gauntface:
cc @wibblymat This should address your feedback.

The input for the original PR accepts Promises too, just had poorly worded error messages. It not accepts Requests and the validation of the input is neatened up a bit.

gauntface:
@wibblymat @jeffposnick 

Had some issues with the tests across browsers - timeout issues. But fixes those and got tests running again in FF (See: https://bugzilla.mozilla.org/show_bug.cgi?id=1249351).

FF Beta isn't working on my local machine at the moment so I cant see how to fix those pending problems.

wibblymat:
Sweet! LGTM

PR #96, opened 2016-02-24
Fixes #87. Adds tests for post method

### LINKED ISSUE #87: Matching on origins should go through all origins
Imagine a toolbox config of the following:

    self.toolbox.router.get('/origin-option-string', function() {
      return new Response('/origin-option-string');
    }, {origin: 'developers.google.com'});

    // This fails because the previous origin matches.
    self.toolbox.router.get('/https-only-string',
      function() {
        return new Response('/https-only-string');
      }, {origin: 'https://developers.google.com'});

What will happen is /https-only-string will never match because toolbox will find the first matching origin (in this case developers.google.com), then tests requests against that set of routes.

gauntface:
@wibblymat @jeffposnick 

This addressed the issues raised here: https://github.com/GoogleChrome/sw-toolbox/issues/87

Tested locally and everything worked as expected.

This makes re-use of the get request tests by passing the desired method string to the sw and test fetch request to match everything up.


gauntface:
I disabled FF beta from the tests since some of the fixes to the service worker API won't land for a week or so and I find myself ignoring errors at the moment.

test/data/router-methods-helper.js
```diff
@@ -0,0 +1,9 @@
+self.getMethodToTest = () => {
+  const methodRegex = /\?method=(\w*)/;
+  const methodRegexResult = methodRegex.exec(location.search);
```
jeffposnick: It would be nice to use https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams instead, but I guess we need to wait a bit for Chrome 49 to make it to stable.

test/data/router-methods-helper.js
```diff
@@ -0,0 +1,9 @@
+self.getMethodToTest = () => {
+  const methodRegex = /\?method=(\w*)/;
```
jeffposnick: Could you switch this to `/[?&]method=(\w+)/` to ensure that it would work if `method` isn't the first parameter, and reinforce the fact that the value can't be empty (`+` vs. `*`).

jeffposnick:
Thanks for fixing #87, and for the additional tests!

(I think the underlying issue reinforces that can sometimes be cleaner to use RegExps instead of splitting up origins and routes.)

test/data/router-methods-helper.js
gauntface: Good catch.

test/data/router-methods-helper.js
gauntface: yeah - realistically I don't expect this test to get to a point where it will need more than this so I'd be inclined to leave it with regex for now.

gauntface:
Yeah origin in the regex is a preferable option, but at the same time, express routes are much faster to work with.

jeffposnick:
:+1:

jeffposnick:
Given that this does have a non-zero chance for breakage due to either folks relying on the old behavior or due to some other unintended consequence, this should probably be rolled out in at least a minor version number bump, rather than a patch bump.

gauntface:
Do you want to file a PR for the minor version bump? Don't know how you or Mat manage releases for these libraries.

jeffposnick:
@wibblymat is the only one with publish rights to `npm`. Cutting a new release for `bower` is a matter of tagging a new version in GitHub.

gauntface:
Awesome sauce. so does anything need to be done or is it just a case of baring in mind? PR suggestion was to change the package.json version

wibblymat:
#90 means the next release needs a major version bump anyway. I'm OOO tomorrow but will do a release on Monday.

(I'd do it now, but it's been a long day and I want to give myself the best chance to not mess it up like I usually do :D)

Side note: Now that npm allows teams we should get all of our stuff owned by a team rather than individuals.

PR #101, opened 2016-03-02
Fixes #100 Cannot read property 'get' of undefined in lib/router.js

### LINKED ISSUE #100: [SW Router] Uncaught TypeError: Cannot read property 'get' of undefined
at https://github.com/GoogleChrome/sw-toolbox/blob/master/lib/router.js#L108

This error occurs for URLs which are not defined with toolbox.router. i.e. all unhandled URLs for which  `methodMap = [undefined]`

adityapunjani:
Brute force fix by adding null check for methodMap


adityapunjani:
cc @jeffposnick

wibblymat:
Thanks Aditya!

It seems a little screwy that methodMaps should ever contain a null value in the first place, but coding defensively doesn't seem like a bad idea anyway.

I think the "real" bug is on L98, as that line means that the array passed to _match will always have one entry, even if there are no matching routes. I'll change that in a follow-up.

adityapunjani:
Thanks for the quick revert @wibblymat

PR #102, opened 2016-03-02
Follow up to #101. Ensure that the `RegExp` origin always exists.

wibblymat:
R: @gauntface @jeffposnick

### DIFF
--- a/lib/router.js
+++ b/lib/router.js
@@ -39,6 +39,8 @@ var keyMatch = function(map, string) {
 
 var Router = function() {
   this.routes = new Map();
+  // Create the dummy origin for RegExp-based routes
+  this.routes.set(RegExp, new Map());
   this.default = null;
 };
 


gauntface:
Sorry, really struggling to see what you are doing here.

Are you basically saying that the origin is the RegExp type? If so this feels like a hacky approach and I'm not entirely sure what the problem is.

Maybe a test case would make it clear what the problem is.

jeffposnick:
I think the confusion might be due to the fact that the identifier `RegExp` is used as the `Map` key that holds all the regular expression routes. See https://github.com/GoogleChrome/sw-toolbox/pull/67 It's confusing, but allowed me to add in regular expression routes without drastically rewriting the existing code.

Given that context, this looks good to me, but I agree with @gauntface that added a test would be a good idea.

gauntface:
What is the problem that this solves?

wibblymat:
The problem is that we [rely on the key already existing](https://github.com/GoogleChrome/sw-toolbox/blob/master/lib/router.js#L98), which until recently meant that you got errors if you didn't have any RegExp routes defined. A check was added further down the line to make sure that the errors stopped, but to prevent any future mishaps I wanted us to not have that problem in the first place.

The simplest way was to ensure that the key really did always exist.

I agree that it is confusing to see the `RegExp` constructor itself used as the key in the map, but there already is [a comment explaining it](https://github.com/GoogleChrome/sw-toolbox/blob/master/lib/router.js#L56) and I'm not sure what a better solution would look like.

wibblymat:
Sorry, to clarify, the error situation was:

1. You have no RegExp routes defined
2. You have no default route
3. A fetch request comes in for a route that doesn't match any Express-style routes

I'll add a test for this.

PR #103, opened 2016-03-03
Adds all methods to tests

gauntface:
and test all methods for 'any' route. Also simplifies the automated test suite - will hopefully work cross platform a little better.

cc @jeffposnick @wibblymat @addyosmani


wibblymat:
:+1:

PR #107, opened 2016-03-30
Moving over to using the sw-testing-helpers + Badges

gauntface:
cc @jpmedley @jeffposnick 

This should get the tests up and running. Slightly concerned that FF Beta is failing at the moment: https://travis-ci.org/GoogleChrome/sw-toolbox/builds/119587028

We've also lost chrome-beta support at the moment due to unmet dependencies when I last tried to install chrome-beta on travis.


.travis.yml
```diff
@@ -1,18 +1,32 @@
-sudo: false
+sudo: true
```
jeffposnick: Is there any way that we could avoid setting this? It means that we're forced to use Travis's older, slower infrastructure that doesn't support caching: https://docs.travis-ci.com/user/migrating-from-legacy/

gauntface: Installing repos to sources list or installing deb's requires sudo. We can revert back once this is fixed: https://github.com/travis-ci/apt-source-whitelist/issues/256

jeffposnick:
:+1: for the actual changes.

Before you merge, though, could you look into what the difference is in your PR's version of the `sw-toolbox.js` and `sw-toolbox.map.json` files vs. what's currently checked into master? The files are minified/large, so the GitHub UI doesn't make it easy to figure out what changed. Since all your changes are for test and Travis code, and that shouldn't make it into in the built library, I'm not sure why there would be any difference at all.

(Also, there's a merge conflict with master that needs to be resolved.)

gauntface:
Thanks for the review Jeff. Changes were accidental (Was debugging an issue between FF and Chrome) Reverted those files and updated with Master. Will merge once tests pass if that's ok with you.

bstn362436:
You fixed it ty

On Wednesday, March 30, 2016, Matthew Gaunt <notifications@github.com>
wrote:

> Merged #107 <https://github.com/GoogleChrome/sw-toolbox/pull/107>.
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/GoogleChrome/sw-toolbox/pull/107#event-608935827>
>


-- 
Sent from Gmail Mobile

Issue #75, opened 2016-01-14
toolbox.precache should not cache invalid files


gauntface:
At the moment toolbox.precache will cache a 404 response.

My initial assumption this in unexpected default behavior (and not something I think would be desirable in any situation).

Would allowing 20x status be a better option?

The concern this raises is what is the behavior when a 404 is encountered? I would expect the service worker install to fail.

jeffposnick:
The background on the current behavior is covered in this issue from the old repo: https://github.com/wibblymat/shed/issues/5

The current configuration is at https://github.com/GoogleChrome/sw-toolbox/blob/master/lib/options.js#L39 and developers could technically override the defaults using whatever `RegExp` they'd prefer, though I'm sure there's no one who actually does that.

My vote is to keep the current behavior, but I can definitely see the justification for the other side.

gauntface:
That bug is in reference to networkFirst behaviour, not the precache behaviour.

The obvious major downside of this is the progressive web app story where you may need and rely on a file that doesn't get cached. In that scenario I'd expect the sw to fail to install.

Appreciate it can be overriden but I don't think this default behaviour is good or expected.

cc @jakearchibald because I know there was discussion around behaviour or cache.addAll and caching 404s.

wibblymat:
Overall I think I'm in favour of being stricter in the precache. A successful install should mean that the app is ready for offline. However, playing Devil's Advocate, here are some cases where the current behaviour is possibly better:

- Caching a custom 404 page. Maybe if you have routes for all of the real URLs on your site you could have the default route just serve up a 404 without checking the server. Precache would fail when it successfully fetched the 404 page if it's served with the correct status code.
- To avoid masking a problem. Let's say that your list of precache resources has a typo in it for some resource. With the proposed behaviour none of your existing users would be able to upgrade, but they also wouldn't hit any errors. It might be some time to even notice that users don't have the latest version. By precaching the 404 you will get bug reports from users much quicker.

gauntface:
# 404

Caching 404 pages is an interesting use case, I'd see two options:
1. Developer creates an endpoint of a query parameter that returns 404 with a 200 status code
2. precache has an additional method that allows the developer to override a valid or invalid response.

This could look something like:

    swtoolbox.precache([
        ' /valid',
        '/my-404'
    ], (asset, response, isValid) => {
        if (asset == '/my-404') {
            return true;
         }
         return isValid; 
    });

[...]

PR #112, opened 2016-04-07
Use NPM Versioning for Dependencies

nickiaconis:
Use NPM versioning for the serviceworker-cache-polyfill dependency instead of github reference.

Our tooling disallows us from using dependencies specified like "coonsta/cache-polyfill", so we can't use sw-toolbox until the dependency specification is fixed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   "main": "lib/sw-toolbox.js",
   "repository": "https://github.com/GoogleChrome/sw-toolbox",
   "dependencies": {
-    "serviceworker-cache-polyfill": "coonsta/cache-polyfill",
+    "serviceworker-cache-polyfill": "^3.0.0",
     "path-to-regexp": "^1.0.1"
   },
   "devDependencies": {


nickiaconis:
I signed it!

gauntface:
Thanks for the PR @nickiaconis - I don't think the polyfill is actually needed anymore so would be worth just removing the dependency all together. What do you think?

NekR:
@gauntface if you are still planning to use `addAll()` then you still need polyfill, see this: https://github.com/GoogleChrome/sw-toolbox/issues/75#issuecomment-206542128

So basically, `addAll()` got new algorithm an polyfill fixes that for older implementations.

NekR:
Also seems that @jakearchibald didn't publish new version to NPM yet.

nickiaconis:
@NekR @jakearchibald Is there a timeframe for the next NPM release of cache-polyfill? Could we not still merge this, since it will pick up the next version once it is published?

NekR:
@nickiaconis next version is 4, so it won't be picked because of `^`.

Currently, `sw-toolbox` points to the repo of polyfill, so this means that it's using latest version (`4.0.0`) and that there is no sense to downgrade it to `3.0.0`. Only thing what we can do is wait for polyfill update on NPm and then modify this PR to point to `4.0.0`.

gauntface:
I'm going to merge this. I understand the changes in the polyfill won't make it to sw-toolbox until the update, but it's bad that we aren't using the npm version at the moment.

@NekR it would probably be worth discussing on the polyfill if the deprecation notice on the library is worth removing given it sounds like it may still be useful

NekR:
@gauntface yep, I created an issue yesterday: https://github.com/dominiccooney/cache-polyfill/issues/21

NekR:
v4 is on NPM now.

addyosmani:
👍

gauntface:
I'm inclined to remove the library all together. I don't think it serves
any purpose.

NekR:
@gauntface depends on what you need. I do not use `addAll()` at all, I use this function which does the same as polyfill and is much smaller: 

```js
function addAllNormalized(cache, requests) {
  return Promise.all(requests.map((request) => fetch(request))).then((responses) => {
    if (responses.some(response => !response.ok)) {
      return Promise.reject(new Error('Response not supported'));
    }

    const addAll = responses.map((response, i) => {
      return cache.put(requests[i], response)
    });

    return Promise.all(addAll);
  });
}
```

PR #115, opened 2016-04-15
Adding tests for routing mechanisms

gauntface:
@jeffposnick @addyosmani @wibblymat if any of you are free for some reviewing time.

I think this is everything that is needed to match the qunit tests. in the 'tests' directory.


test/browser-tests/cache-first/cache-first.js
```diff
@@ -0,0 +1,103 @@
...
+    })
+    .then(response => {
+      return response.text();
+    })
+    .then(responseText => {
+      responseText.trim().should.equal(TEST_INPUT);
+    });
+  });
+
+  it('should retrieve the first value from the network and then the alteredcache', function() {
```
addyosmani: `alteredcache` -> `altered cache`?

test/browser-tests/cache-first/cache-first.js
```diff
@@ -0,0 +1,103 @@
...
+
+'use strict';
+
+var swUtils = window.goog.SWHelper;
+
+describe('Test toolbox.cacheFirst', function() {
+  const serviceWorkersFolder = '/test/browser-tests/cache-first/serviceworkers';
+
+  it('should retrieve the first value from the cache and not update it', function() {
+    var iframe;
```
addyosmani: Friendly reminder to switch out these `var` statements since we're using `const` and `let` elsewhere.

addyosmani:
@gauntface I had a look over this again and if we can get the comments above and breaking build issues resolved on this one I'd be happy to land. It would still be great to get @wibblymat's eyes on this if he has a chance.

gauntface:
Thanks for the review @addyosmani 

Hopefully fixed the tests (although I'm not happy with the fix). Essentially Firefox was pointing out an issue where toolbox was returning a response to the browser before it had updated the cache and as a result the test was querying the cache immediately before the value had been added. I've just added some timeouts but that is still likely to cause a race condition. Need to think of a better approach but right now not sure what the solution would be.

addyosmani:
I forgot to hit LGTM. So..LGTM anyway!

PR #119, opened 2016-04-20
Move docs into subfolder.

### LINKED ISSUE #118: Move documentation from README
Currently it is easy to miss the API documentation in the README. While it is a huge chunk of content, it also isn't exactly clear that is the docs either.

We should move the API docs under a `doc` folder and cross-link that. Further we may want to take the usage section and move that under a `usage.md` file of the docs. This way the README is more precise to inform people about how the project helps them, how to include it, and what the license is.

I am going to try and have a PR made this afternoon to address this. It is critical that the documentation is very easy to find and scan over without the other content getting in the way.

Garbee:
Fixes #118


addyosmani:
LGTM

PR #120, opened 2016-04-20
Adding support for auto-building docs on release + travis

gauntface:
@Garbee would appreciate your thoughts on this

cc @jeffposnick as an FYI


Garbee:
just ran and checked out the output. LGTM for an initial start.

Minor nitpick on organization... usage/API should be under a "docs" heading while the recipes would fit best under a "Demos" heading leaving room for more later.

gauntface:
This seems like a major downside of jsdoc (unless you know of a trick to
achieve this, but from looking the docs for jsdoc, everything has to be a
tutorial, within that we can maybe have docs and demos.

Garbee:
May want to peek at [esdoc](https://esdoc.org/tutorial.html#create-config-file). It supports multiple "manual" pages that would fit the needs here.

However, that can be updated later. For now this works so I'm good with it.

gauntface:
Cheers @Garbee

addyosmani:
This LGTM. I'm down with us switching to ESDoc where needed as we evaluated it internally for DevRel projects and liked it too.

PR #126, opened 2016-04-27
Build / Publishing Improvements. Updates sw-testing-helpers.

gauntface:

At the moment sw-toolbox has a compiled version in the master repo, this is pretty bad practice and would like to kill it.

This PR makes use of a publishing script that will build the project, run tests against the build, build the docs, bundle everything for a release and publish to NPM, with this, we can now remove the static build in the master repo.

If developers still need git access to also get the pre-build versions the 'release' branch which the publish script creates can be used (or tagged releases).

This also updates to the latest sw-testing-helpers which

- Fixes TestServer binding to localhost rather than responding to external requests
- Access to all useful classes without requiring from node_modules
- Re-usable release and doc publishing scripts
- Doc template is shared and sorts revisioned directories correctly.


test/automated-suite.js
```diff
@@ -24,7 +24,7 @@
 
 require('chai').should();
 const path = require('path');
-const mochaUtils = require('../node_modules/sw-testing-helpers/src/mocha/utils.js');
+const mochaUtils = require('sw-testing-helpers').mochaUtils;
```
wibblymat: FYI - what you were doing before could also have been done cleaner as `require('sw-testing-helpers/src/mocha/utils')`. Require will find things relative to installed modules. Exporting the things that are needed is even better though, of course!

PR #128, opened 2016-04-27
New sw-testing-helpers + Remove build artefacts.

gauntface:

At the moment sw-toolbox has a compiled version in the master repo, this is pretty bad practice and would like to kill it.

This PR makes use of a publishing script that will build the project, run tests against the build, build the docs, bundle everything for a release and publish to NPM, with this, we can now remove the static build in the master repo.

If developers still need git access to also get the pre-build versions the 'release' branch which the publish script creates can be used (or tagged releases).

This also updates to the latest sw-testing-helpers which

Fixes TestServer binding to localhost rather than responding to external requests
Access to all useful classes without requiring from node_modules
Re-usable release and doc publishing scripts
Doc template is shared and sorts revisioned directories correctly.


wibblymat:
This LGTM, assuming you get Travis working :)

gauntface:
Should be fixed now - turns out spacing within an if statement is very important for Travis, otherwise it dies quietly in the background and github makes it looks like it's still waiting :-(

addyosmani:
LGTM. Waiting for Travis checks to complete.

gauntface:
I'm seeing this Firefox Beta problem across a few different projects - happy to accept given that tests are passing in FF stable and both chromes?

addyosmani:
Happy to accept. Assume we're going to look at how to tackle the beta breakage beyond this project.

Nice one.

PR #130, opened 2016-05-03
clean up redundant caches. domain global though

donlion:
On several projects, where I've used sw-toolbox, I'm using versioning numbers as well to invalidate all my files at once - when it's needed of course.
Only problem is, that I ended up having multiple redundant caches, that basically just filled up space.

The challenge in removing redundant caches is the fact, that it's removing it form the whole domain.
Meaning if the website has a subsite under /serviceworker/is/the/best using a scoped serviceworker, their cache might all of a sudden not be available.

I'm not sure how to solve this yet, but I really badly needed the clean up and thought I might as well just implement it my self :)

How should we go around this?

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -132,9 +132,29 @@ function renameCache(source, destination, options) {
   });
 }
 
+
+function renameCacheAndCleanUp(source, destination, options) {
+  debug('Renaming cache: [' + source + '] to [' + destination + ']', options);
+
+  return caches.keys().then(function(cacheNames) {
+    Promise.all(
+
+      cacheNames.map(function(cacheName) {
+        if (destination !== cacheName && source !== cacheName) {
+          return caches.delete(cacheName);
+        }
+      })
+
+    ).then(function() {
+      return renameCache(source, destination, options);
+    });
+  });
+}
+
 module.exports = {
   debug: debug,
   fetchAndCache: fetchAndCache,
   openCache: openCache,
-  renameCache: renameCache
+  renameCache: renameCache,
+  renameCacheAndCleanUp: renameCacheAndCleanUp
 };
--- a/lib/sw-toolbox.js
+++ b/lib/sw-toolbox.js
@@ -73,7 +73,7 @@ self.addEventListener('install', function(event) {
 self.addEventListener('activate', function(event) {
   helpers.debug('activate event fired');
   var inactiveCache = options.cache.name + '$$$inactive$$$';
-  event.waitUntil(helpers.renameCache(inactiveCache, options.cache.name));
+  event.waitUntil(helpers.renameCacheAndCleanUp(inactiveCache, options.cache.name));
 });
 
 // Fetch


donlion:
Signed, sealed, delivered

gauntface:
Thanks for this @donlion, could you add some tests to make sure this works and doesn't regress.

I'm not sure if the renameCache method needs to be exposed, I'd expect the renameCacheAndCleanUp to be the only methods OR have two methods renameCache() cleanUpCaches().

@wibblymat and @jeffposnick what are your thoughts on this? I hit this problem last night on my own site.

donlion:
Yeah. Reason why I kept renameCache and renameCahceAndCleanUp is for the domain-global "danger".
Say I've got domain.com as my main website and want to launch a web-app under the directory domain.com/web-app. If the users landed on domain.com and it cleaned up the cache, the cache for my web-app would be gone as well.
This is obviously a set example, but it could happen.

How should we go around it?
We could "namespace" the cache name to the service-workers scope and only remove caches within the current scope. What do you think?

addyosmani:
> I'd expect the renameCacheAndCleanUp to be the only methods OR have two methods renameCache() cleanUpCaches().

I concur.

> @wibblymat and @jeffposnick what are your thoughts on this? I hit this problem last night on my own site.

Friendly ping Mat and Jeff for your take on this before we ask @donlion for tests.

jeffposnick:
The default cache name [already contains](https://github.com/GoogleChrome/sw-toolbox/blob/4d172db13087d5727c6a3fe73a9341e6f2c4b5e4/lib/options.js#L29) the current service worker's scope:

```
name: '$$$toolbox-cache$$$' + scope + '$$$'
```

I may be missing something subtle here, but it doesn't sound like multiple service workers' usage of `sw-toolbox` would introduce a conflict, since they each should have their own scope.

lib/helpers.js
```diff
@@ -132,9 +132,29 @@ function renameCache(source, destination, options) {
...
+
+function renameCacheAndCleanUp(source, destination, options) {
+  debug('Renaming cache: [' + source + '] to [' + destination + ']', options);
+
+  return caches.keys().then(function(cacheNames) {
+    Promise.all(
+
+      cacheNames.map(function(cacheName) {
+        if (destination !== cacheName && source !== cacheName) {
+          return caches.delete(cacheName);
```
jeffposnick: This seems dangerous!

If a user is either managing a cache outside of `sw-toolbox` (e.g. via `sw-precache` or just vanilla Cache Storage API calls) or using `sw-toolbox` but with explicitly named caches (e.g. in [this example](https://github.com/GoogleChrome/sw-toolbox/blob/master/recipes/cache-expiration-options/service-worker.js#L18)), then those caches would end up being deleted by this code.

wibblymat:
Thanks for the PR!

I am currently -1 on this. If you use the default cache name then toolbox already manages it for you, and you don't have to clean anything up. If you create your *own* caches, then you should do your own tidying up too.

Maybe we could come up with some sort of optional helper, though, that makes tidying up custom caches  easier. Maybe something like `expireCaches(prefix, suffix)` that will, during `activate`, delete any cache that starts with `prefix` but does not end with `suffix`. This should help with versioned caches where the prefix will contain the scope and the suffix will be the current version. So if you have a cache called `scope + '-images-' + version`, then when you upgrade from version 4 to version 5 the version 4 cache will be deleted for you.

Does that make sense? Would it actually solve your problem?

addyosmani:
Thanks for the initial review of the idea, @wibblymat and @jeffposnick. My two cents are that this PR would introduce sufficiently dangerous behavior that it's not worth pursuing in the current form. We do however greatly appreciate the contribution.  

> If a user is either managing a cache outside of sw-toolbox (e.g. via sw-precache or just vanilla Cache Storage API calls) or using sw-toolbox but with explicitly named caches (e.g. in this example), then those caches would end up being deleted by this code.

This is my concern too thus a -1 I'm afraid.

> Maybe we could come up with some sort of optional helper, though, that makes tidying up custom caches easier. Maybe something like expireCaches(prefix, suffix) that will, during activate, delete any cache that starts with prefix but does not end with suffix. This should help with versioned caches where the prefix will contain the scope and the suffix will be the current version. So if you have a cache called scope + '-images-' + version, then when you upgrade from version 4 to version 5 the version 4 cache will be deleted for you.

I think this might be worth moving into a new issue. I'll file one for us to explore whether there's more that can be done to ease the process of tidying up caches.

PR #145, opened 2016-06-09
Fixes #143

### LINKED ISSUE #143: Restore previous top-level sw-toolbox.js
It broke `sw-precache`, as https://github.com/GoogleChrome/sw-precache/blob/master/lib/sw-precache.js#L184 depended on it existing.

When you do remove it, could it be as part of a major `sw-toolbox` semver release?

gauntface:
@jeffposnick @addyosmani @wibblymat You ok with this?

Would a patch fix for this be ok?

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -55,8 +55,8 @@ gulp.task('build', function() {
 
   bundler.plugin('browserify-header');
   bundler.plugin('minifyify', {
-    map: 'sw-toolbox.map.json',
-    output: 'sw-toolbox.map.json'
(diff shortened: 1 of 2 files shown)


addyosmani:
LGTM. This is in line with patch release to fix the breaking builds elsewhere and major release to/if we change the path.

gauntface:
Tests for Chrome stable and beta have passed so going to merge this and push a patch release. They've failed in Firefox but thats a selenium issue.

wibblymat:
LGTM - I'll do a new release (v3.2.1) now

PR #154, opened 2016-06-29
Fixes #3

### LINKED ISSUE #3: Warn if routes overwrite each other
If I have a route for `:foo/:bar` and another for `:baz/:quux` then they compile to the same regex, and one will overwrite the other.

This should cause an error or warning in the console.

wibblymat:
R: @gauntface @jeffposnick

This is kind of trivial, but it happened to be the oldest open issue when I started reviewing them all.

I tried to work out how to write a test for this in the new framework, but there doesn't seem to be anything equivalent.

### DIFF
--- a/lib/router.js
+++ b/lib/router.js
@@ -16,6 +16,7 @@
 'use strict';
 
 var Route = require('./route');
+var helpers = require('./helpers');
 
 function regexEscape(s) {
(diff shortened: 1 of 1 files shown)


addyosmani:
LGTM, but will defer to @gauntface or @jeffposnick for a comment about tests.

PR #159, opened 2016-07-08
NetworkFirst: return responses on client errors

gierschv:
When a success response was cached and the newly returned response
was a client error (e.g. 402, 403), this error was ignored and the
cache was returned. The new behavior is to directly return those
client errors and avoiding to fallback on the existing cache.

### DIFF
--- a/lib/options.js
+++ b/lib/options.js
@@ -36,5 +36,9 @@ module.exports = {
   // A regular expression to apply to HTTP response codes. Codes that match
   // will be considered successes, while others will not, and will not be
   // cached.
-  successResponses: /^0|([123]\d\d)|(40[14567])|410$/
+  successResponses: /^0|([123]\d\d)|(40[14567])|410$/,
+  // HTTP Errors codes that match will be returned to the client and will
+  // not fallback to a cached response. For example we had a 200 cached and
+  // our `toolbox.networkFirst` now returns 403.
+  errorsResponsesWithoutCacheFallback: /^4\d\d$/
 };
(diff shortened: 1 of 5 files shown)


NekR:
Just curious, why the new behavior is correct and the old one not?

gierschv:
@NekR Because the old behavior considers client errors as a network errors. I believe this should not be the case, but if the old behavior needs to be conserved as-is, I'd like to see this PR as an option.

NekR:
I don't say I shouldn't be landed or something. It's just interesting to me why it's the correct behavior in your mind and which use cases you had where old behavior didn't work (I can imagine few, just would be good to know real example).

gierschv:
@NekR Sure, for example we cache some GET API responses with the `networkFirst` strategy, so an app can be used without network and we are sure the data are always fresh. Some day, the user doesn't have access to the cached resource anymore, for example because of his session (e.g. error 403), in the old behavior, the SW will keep returning a 200 and never consider the HTTP Client error.

NekR:
I see, thanks for info :-)

jeffposnick:
I'm doing some housekeeping on older PRs. Apologies for not getting to this sooner.

`successResponses` is configurable per-route. If there's a set of HTTP response codes that you want to be considered "final" for a given route, like HTTP 403, then I'd recommend that you pass in a regular expression that matches those responses in that route's configuration.

toolbox.router.get('/path/to/api', toolbox.networkFirst, {
  successResponses: /^0|([123]\d\d)|(40[1234567])|410$/
});

I hesitate to do anything that changes the default behavior of something as core as the `networkFirst` handler, given how mature `sw-toolbox` is at this point.

PR #165, opened 2016-07-28
implements oldest first cache expiry option

wheresrhys:
Hello,

This PR implements 'first in first out' cache expiry where maxEntries is set.

Personally I think this would make more sense as the default behaviour, though for backwards compatibility this PR, offering an easy way to configure via global options, seems like a less controversial offering

If you're happy with this in principle I'll add tests and docs

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -84,6 +84,7 @@ function cleanupCache(request, cache, cacheOptions) {
   var requestUrl = request.url;
   var maxAgeSeconds = cacheOptions.maxAgeSeconds;
   var maxEntries = cacheOptions.maxEntries;
+  var expireOldestFirst = cacheOptions.expireOldestFirst;
   var cacheName = cacheOptions.name;
 
   var now = Date.now();
@@ -93,7 +94,7 @@ function cleanupCache(request, cache, cacheOptions) {
   return idbCacheExpiration.getDb(cacheName).then(function(db) {
     return idbCacheExpiration.setTimestampForUrl(db, requestUrl, now);
   }).then(function(db) {
-    return idbCacheExpiration.expireEntries(db, maxEntries, maxAgeSeconds, now);
+    return idbCacheExpiration.expireEntries(db, maxEntries, maxAgeSeconds, expireOldestFirst, now);
   }).then(function(urlsToDelete) {
     debug('Successfully updated IDB.');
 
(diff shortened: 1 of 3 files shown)


jeffposnick:
Hello, @wheresrhys!

Currently, when `maxEntries` is specified, there's a least-recently used cache expiration policy imposed. (Every time a `Response` is read/written to the cache, the corresponding entry in `IndexedDB` has its timestamp updated.)

While I haven't actually run your code, it seems to me that swapping the `direction` value will just serve to implement a most-recently used cache expiration policy, rather than a first-in, first-out policy.

Taking a step back, do you have a specific use case in mind for which first-in, first-out provides noticeable benefits vs. least-recently used? I don't mean to imply that it's an invalid feature, but LRU does end up being a good fit for many use cases.

Also, if we do go ahead with this, please ensure that all of the existing tests [pass](https://travis-ci.org/GoogleChrome/sw-toolbox/builds/148061359), and we would greatly appreciate additional tests that cover the new functionality.

addyosmani:
It's been a few months since we've heard back on this one, but I would also be interested in understanding the use-case here before we move ahead with trying to iterate on the idea. Any thoughts/interest in helping us out with your insights, @wheresrhys?

addyosmani:
Closing this one for now. If you get a chance to reping us on this @wheresrhys we'd be happy to chat more!

wheresrhys:
Apologies for slow reply (working on SW at work, but emails fro this repo go to my personal account)

There was a bit of misdiagnosis of caching issues we were having with sw-toolbox, and working on this PR helped identify them in the end, and I did mean to close it afterwards. One bit of unexpected behaviour with sw-toolbox's cache though is that it is a soft purge (i.e. the cache is purged and repopulated lazily after the request is responded to), which isn't suitable behaviour for all our needs. Though again it may have been a misdiagnosis of the problem (I'm working on getting a lot more tests up and running for our SW now)

jeffposnick:
If you find the need to get notified when the "soft purge" happens, the changes in https://github.com/GoogleChrome/sw-toolbox/pull/174 might be of interest to you.

wheresrhys:
That does look interesting, but it was more for instances where e.g. we cache an index page and then the user doesn't revisit for a week - in this case we would want a hard purge so that if a user hasn't revisited within, say, an hour, then the cached resource will be discarded without being used.

PR #172, opened 2016-08-15
Update skip-and-claim.js

gauntface:
Use event.waitUntil();

### DIFF
--- a/test/data/skip-and-claim.js
+++ b/test/data/skip-and-claim.js
@@ -1,4 +1,4 @@
-self.addEventListener('install', function() {
+self.addEventListener('install', function(event) {
   event.waitUntil(self.skipWaiting());
 });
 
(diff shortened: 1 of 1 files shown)


NekR:
Is there a downside of not using `event.waitUntil()`? I saw some people not using it other libraries.

gauntface:
The browser implementations basically up ambiguities.

Say browser X was super slow to run skipWaiting, but had a very aggressive service worker kill policy, not using event.waitUntil will allow the browser to kill the service worker before skipWaiting is executed and finished.

NekR:
Thanks, that makes sense :-)

jeffposnick:
FWIW, I don't think this change was needed; as [per the spec](https://slightlyoff.github.io/ServiceWorker/spec/service_worker/#service-worker-global-scope-skipwaiting), the promise returned by `skipWaiting()` is resolved in parallel with the activate algorithm being started, which is to say that it doesn't wait until the activation completes before it resolves.

I don't *think* that even a very aggressive browser would kill the service worker in the middle of the activation algorithm.

(I used to add the promise returned from `skipWaiting()` and `clients.claim()` to the `event.waitUntil()` chain in all the code I wrote until @jakearchibald pointed out that this didn't actually accomplish anything.)

PR #174, opened 2016-08-25
Adds new notifyOnCacheUpdate cache option

### LINKED ISSUE #116: Opt-in to postMessage notifications when fastest updates the cached content
The `toolbox.fastest` strategy provides a convenient way of returning a previously cached `Response` immediately, while at the same time `fetch()`ing a potentially newer copy from the network.

However, if the network request did result in an updated copy being retrieved, the page that requested the resource does not currently receive any notification that there's a newer `Response` available.

We can add in a new option that would allow developers to opt-in to using `ServiceWorkerClient.postMessage()` to trigger an event handler on the controlled page, letting it know that the cached entry is out of date.

jeffposnick:
R: @addyosmani @gauntface @wibblymat 

This adds in support for a new `notifyOnCacheUpdate` option, which is most useful alongside a `fastest` cache strategy. It allows developers to opt-in to receiving a `message`-based notification on their `navigator.serviceWorker` interface whenever there's an update to the underlying cached resource, letting them know what the URL is that was updated.

This attempts to determine whether or not a resource was updated by comparing the `last-modified`, `etag`, and `content-length` response headers. Since response objects can be arbitrarily large, it seemed safer not to attempt to do a byte-by-byte comparison of the actual response bodies, and I think that's a reasonable workaround.

I'm adding an additional test, documentation, and a new sample in the `recipes/` directory to accompany this new feature.

Fixes #116


TimvdLippe:
I could imagine that using this option for json files with data, you could potentially handle it right away without a reload. Would an extra option (or default behavior) to send the actual content of the file in the message be possible?

A usecase would be the offline-first chat-app Jake Archibald built. This app could be notified with "Hey, these are all the new messages, enjoy!" and immediately push them to the user without having him/her to refresh the page.

jeffposnick:
If the page knows that it's going to use the new content, it could call `caches.match(event.data.url).then(response => response.json())` from the context of the controlled page, since `caches` is exposed on the `window` global.

But since not everyone would know to do that, and since that's a good use case, I'll update the sample with a comment to let developers know.

jeffposnick:
I just pushed a few changes to make what you were suggesting easier, @TimvdLippe. The `MessageEvent`'s `data` now has the name of the cache, so you can explicitly open that cache and match against it, instead of having to do a less efficient `caches.match()`. I also re-ordered the timing of the `postMessage()`, so that it happens after the `cache.put()` promise has resolved, ensuring that you'll actually get the updated resource if you call `.match()` from the controlled page.

Thanks for the suggestion!

TimvdLippe:
No problem and glad to see the improvements! :tada:

addyosmani:
Although we're still seeing some flakiness with Travis and I've pinged Matt about taking a look at what we can do to move this back along there, the changes here LGTM. Appreciate the addition of a recipe entry as that was something I was going to suggest initially. Otherwise this is all good to land. Nice work, Jeff!

PR #187, opened 2016-09-22
Refactors the tests of cache expiration options.

jeffposnick:
R: @gauntface
CC: @addyosmani 

This fixes up a test that was flaky due to the need to clear out IDB in between invocations, which didn't always succeed due to the DB instance being held open in the service worker.

It's a refactoring of the logic so that it's no longer necessary to clear out IDB, since there's a separate DB instance used for each individual sub-test (scoped to the service worker registration).

I also took the opportunity to refactor some of the promise chains to make them more readable, and move common logic out to helper functions.


test/browser-tests/options/options.js
```diff
@@ -20,209 +20,117 @@
...
-  });
+  /**
+   * Asserts that the keys in cachedAssets match exactly the list of expected URLs.
+   * @param {Object} cachedAssets The result from a call to swUtils.getAllCachedAssets().
+   * @param {Array.<String>} expectedUrls The expected cache contents.
+   */
+  const assertCacheContents = (cachedAssets, expectedUrls) => {
+    const expectedLength = expectedUrls.length;
+    const cachedUrls = Object.keys(cachedAssets);
+    const filteredUrls = cachedUrls.filter(url => expectedUrls.includes(url));
```
gauntface: Not sure of when filteredUrls != cachedUrls and that would be OK.

Would it make sense to invert the includes check and ensure filteredUrls.length === 0?

jeffposnick: All of the tests in this file assume that the cached contents need to be exactly equal to the expected contents. I.e., if there's something in cached contents that's not in expected, or vice versa, then it's a failure. This logic seemed like the easiest way to check that bijection.

It's not really a general-purpose check that's applicable to other tests, since other tests might be implemented in such a way that extra entries in the cached contents that aren't expected wouldn't be considered a failure.

PR #191, opened 2016-09-22
chrome-extension requests shouldn't trigger the default route

### LINKED ISSUE #171: Default route doesn't check request protocol, picks up chrome-extension://
error with extensions 

Example given during NYC Roadshow

sw-toolbox.js:18 Uncaught (in promise) TypeError: Request scheme 'chrome-extension' is unsupported(…)(anonymous function) @ sw-toolbox.js:18

jeffposnick:
R: @addyosmani @gauntface 

Explicitly check to make sure that the request URL starts with `'http'` before triggering the default route.

Fixes #171

### DIFF
--- a/lib/sw-toolbox.js
+++ b/lib/sw-toolbox.js
@@ -83,7 +83,10 @@ self.addEventListener('fetch', function(event) {
 
   if (handler) {
     event.respondWith(handler(event.request));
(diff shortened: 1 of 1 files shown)


gauntface:
LGTM

PR #192, opened 2016-09-23
Use uglify instead of minifify and fix up sourcemap generation

### LINKED ISSUE #169: Is source map working properly?
Hello, 

I came to know this great tool for service worker. I've learned many thing by this tool. Thank you. :)
But I get to have question while using it.

For debugging I have used source-map that was built. 

First I had a problem because of source map path in sw-toolbox.js
`//# sourceMappingURL=./build/sw-toolbox.map.json`
But It's not big problem to me. I can be solved by remove `build`.
`//# sourceMappingURL=./sw-toolbox.map.json`

But my big problem is **"I could not take a break point" although its source tree appears on dev tools.**

image

I've tried to make break point on `Router.prototype.matchMethod` function (router.js) but the break point is made on fastest.js(53). Some breakpoints are made well. but most of all are not made on proper code.

So I tried another way of making source map by modifying gulpfile. 

I referenced this
https://github.com/gulpjs/gulp/blob/master/docs/recipes/browserify-uglify-sourcemap.md

```js
 var bundler = browserify({
    entries: ['./lib/sw-toolbox.js'],
    standalone: 'toolbox',
    debug: true
  });
  var bundle = function() {
    return bundler
      .bundle()
      .pipe(source('sw-toolbox.js'))
      .pipe(buffer())
      .pipe(sourcemaps.init({loadMaps: true}))
      .pipe(uglify())
        // Add transformation tasks to the pipeline here.
      .pipe(sourcemaps.write('./'))
      .pipe(gulp.dest('./build/'));
  };
  return bundle();
```

It works very well.

What was my problem? How can I use source-map by not modifying gulf file. 
Thank you for your help in advance.

jeffposnick:
R: @addyosmani @gauntface 

Going with `uglify` as recommended in the [`gulp` docs](https://github.com/gulpjs/gulp/blob/master/docs/recipes/browserify-uglify-sourcemap.md) seems to lead to a better-behaving sourcemap when used from Chrome's debugger.

It also leads to a slightly smaller final bundle (15kb vs 17kb with `minifify`).

This also fixes up the URL pointing to the sourcemap file, telling the debugger to look for the sourcemap in the same directory as the `sw-toolbox.js` library, rather than in a `./build/` subdirectory.

I decided to just inline the Apache 2.0 license header that gets prepending to the generated library, but if you think it's too messy that way, I can put it in a standalone file and then call `fs.readFileSync()` to bring it into the build.

Fixes #169

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -15,14 +15,18 @@
 */
 'use strict';
 
-var path = require('path');
 var browserify = require('browserify');
+var buffer = require('vinyl-buffer');
 var eslint = require('gulp-eslint');
 var ghPages = require('gulp-gh-pages');
 var gulp = require('gulp');
+var header = require('gulp-header');
+var path = require('path');
 var source = require('vinyl-source-stream');
+var sourcemaps = require('gulp-sourcemaps');
 var temp = require('temp').track();
 var testServer = require('./test/server/index.js');
+var uglify = require('gulp-uglify');
 
 var buildSources = ['lib/**/*.js'];
 var lintSources = buildSources.concat([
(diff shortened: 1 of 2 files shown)


gauntface:
LGTM and +1 the license header approach, looks clean to me.

PR #193, opened 2016-09-23
Fix some paths in the cache-expiration-options recipe

jeffposnick:
R: @addyosmani @gauntface 

Nothing major here. Just fixing some paths up so that they work by default when run locally.


gauntface:
LGTM

PR #195, opened 2016-09-28
Separate pure and side-effecting code

samertm:
To make it easier to use sw-toolbox as a library, this PR moves all the helper and listener functions out of sw-toolbox.js so they can be imported without setting up the event listeners.

Tests in Chrome 51 pass locally.

### DIFF
--- a/lib/sw-toolbox.js
+++ b/lib/sw-toolbox.js
@@ -15,103 +15,22 @@
 */
 'use strict';
 
-require('serviceworker-cache-polyfill');
+// This is the entrypoint for the sw-toolbox bundle. All code with
+// side effects (e.g. adding event listeners) should be in this file.
+
 var options = require('./options');
 var router = require('./router');
 var helpers = require('./helpers');
 var strategies = require('./strategies');
+var listeners = require('./listeners');
 
 helpers.debug('Service Worker Toolbox is loading');
 
-// Install
-var flatten = function(items) {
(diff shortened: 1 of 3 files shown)


samertm:
I signed it!

lib/sw-toolbox.js
```diff
@@ -121,7 +40,7 @@ module.exports = {
   fastest: strategies.fastest,
   router: router,
   options: options,
-  cache: cache,
-  uncache: uncache,
-  precache: precache
+  cache: helpers.cache,
+  uncache: helpers.uncache,
+  precache: helpers.precache
 };
```
addyosmani: This cleans up the main sw-toolbox entry quite a bit while making it easier to just pull in the helpers. It's also passing our current CI tests. I'm personally fine with this change, but let's see what @jeffposnick and @gauntface think.

jeffposnick:
👍 from my perspective as well.

As an FYI, we're thinking about similar topics in https://github.com/GoogleChrome/sw-helpers/issues/44

I see this PR a move in that direction using the current codebase, which should be continue to be relevant for a while yet. But if you have any feedback about the proposals tracked in that issue, that would end up being the most meaningful longer term.

samertm:
Thanks! I'll take a look at that issue.

samertm:
Friendly ping to merge this in. :)

Issue #196, opened 2016-10-05
Proposal: Pass "event" to strategies


samertm:
Currently, handlers/strategies have access to the following information: request, values, and options

I'm extending sw-toolbox by writing my own strategies, and there are some things I can't do because I don't have access to the event itself. E.g., I want to be able to call `waitUntil` and use the `clientId` in my strategy.

Express passes "request" and "response" objects to every route handler, which makes that framework very flexible. I propose we do something similar with sw-toolbox handlers:

1. Pass "event" as the first argument to handlers.
2. Handlers are responsible for calling "event.respondWith" synchronously if they want to respond to an event.

The modifications needed are minimal and arguably reduce the "magic" of the framework. This is what cacheOnly would look like:
```
function cacheOnly(event, values, options) {
  var request = event.request;
  helpers.debug('Strategy: cache only [' + request.url + ']', options);
  event.respondWith(helpers.openCache(options).then(function(cache) {
    return cache.match(request);
  }));
}
```

And networkOnly could look like this:
```
function networkOnly(event, values, options) {
  helpers.debug('Strategy: network only [' + event.request.url + ']', options);
  // We don't call event.respondWith, so the browser simply makes a request.
  // (Equivalent to: `event.respondWith(fetch(event.request))`). 
}
```

Thoughts?

samertm:
> (Though I'm not sold on handlers taking responsibility for calling event.respondWith().)

My thought there was that it brings sw-toolbox closer to the express framework:

```
app.get('/', function(req, res) {
  res.send("some response");
});
```

The proposal still works without it, but I think it's a familiar pattern for web devs.

samertm:
Re: making a large change like this, we could bump the major version number and add a note in the documentation about the change.

PR #197, opened 2016-10-06
RFC: Pass 'event' instead of 'request'

samertm:
This PR updates request handlers to take `event` instead of `request` as their first argument, as discussed in #196. Request handlers need access to `event` because people that build on top of sw-toolbox may need access to `event.client` and `event.waitUntil`.

I didn't move event.respondWith inside of the event handlers because it seemed potentially error-prone and more verbose, because event.respondWith has to be called synchronously inside of the handler.

Looking at this, I'm not 100% sure it's worth breaking backwards compatibility.

Two alternatives:
- Pass in the event object as the last arg: `handler(request, values, options, event)`
- Add the event object to "options" (which overloads the use of options, which is currently only to override cache options):

```
handler(request, values, options) {
  // use options.event here
```

Thoughts?

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -46,8 +46,8 @@ Some examples of using Express-style routing include:
 ```javascript
 // URL patterns are the same syntax as Express routes
 // (http://expressjs.com/guide/routing.html)
-toolbox.router.get(':foo/index.html', function(request, values) {
-  return new Response('Handled a request for ' + request.url +
+toolbox.router.get(':foo/index.html', function(event, values) {
+  return new Response('Handled a request for ' + event.request.url +
       ', where foo is "' + values.foo + '"');
 });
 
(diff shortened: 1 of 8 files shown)


gauntface:
@samertm sorry that this has been sat without response for so long.

My gut says that we aren't likely to land this as it feels like a big enough change to break people and we'd rather encourage use of the sw-helpers for this kind of use case (sw-helpers is essentially a breakdown of sw-toolbox and sw-precache with the intention of allowing developers to build higher level abstractions on top of it)

addyosmani:
I'm going to echo Matt's comments here and suggest that any backwards compat breaking changes such as this be considered in sw-helpers instead of sw-toolbox. That would greatly help us keep toolbox and precache relatively stable-train solutions for the next quarter while we continue to evolve the next versions of them in the helpers project: https://github.com/googlechrome/sw-helpers

PR #199, opened 2016-10-17
Further tweaks to package.json's files field

jeffposnick:
R: @gauntface 
CC: @addyosmani 

I was taking another look at the publishing process prior to cutting the new release of `sw-toolbox`, and I realized that the output of the `build/` directory is explicitly copied over to the top-level directory via [`npm-publish-scripts`](https://github.com/GoogleChrome/npm-publish-scripts/blob/master/project/create-release-bundle.sh#L26), so including `build/` in the `package.json` `files` field is not necessary.

Instead, I'm explicitly adding in entries for the built `sw-toolbox.js` and corresponding sourcemap, which should hopefully correspond to what `npm-publish-scripts` will copy over.

### DIFF
--- a/docs/recipes/cache-expiration-options/index.html
+++ b/docs/recipes/cache-expiration-options/index.html
@@ -69,6 +69,6 @@
     <div id="results"></div>
 
     <script src="app.js"></script>
-    <script src="../../companion.js" data-service-worker="service-worker.js"></script>
+    <script src="/companion.js" data-service-worker="service-worker.js"></script>
   </body>
 </html>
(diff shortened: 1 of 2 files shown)


gauntface:
LGTM.

For a side note, I'm updating npm-publish-scripts to be more "npm friendly", using npmignore and files instead of it's bundle step. But for the next release carry on with this and we can look at updating npm-publish-scripts in a seperate PR.

addyosmani:
LGTM2. Nice work, @jeffposnick.

PR #206, opened 2016-11-21
Explicitly check the Date: header of a cached response when maxAge is set

### LINKED ISSUE #164: maxAgeSeconds and `fastest`
When used with `fastest` strategy, the `maxAgeSeconds` option somehow becomes peripheral. Because the cache always expires after a response is returned.

When users refresh a page served by `fastest`, existing cache always kicks in first, even that it might already expired.

``` javascript
toolbox.cacheFirst('/mypage', {
    cache: {
        name: 'mypage',
        maxAgeSeconds: 10
    }
});
```

jeffposnick:
R: @adityapunjani @gauntface @addyosmani
CC: @ktmud

My current thinking is that if we release this PR, just use a minor version bump, and don't require opting-in to an additional option to turn it on. While it does change the current behavior, based on #164, it's a change that reflects the behavior that some developers had always assumed to be the case. The net effect is that entries that would have previously been valid when read from the cache are instead expired sooner, but since they're entries that would have been expired after the fact anyway, I don't see it as a breaking change.

Alongside the code review, I'd like to get some confirmation from @adityapunjani & co. that this change works for them prior to merging and cutting a new release.

Fixes #164


gauntface:
This looks ok to me.

lib/helpers.js
```diff
@@ -171,6 +171,33 @@ function validatePrecacheInput(items) {
...
+  // Only bother checking the age of the response if maxAgeSeconds is set.
+  if (maxAgeSeconds) {
+    var dateHeader = response.headers.get('date');
+    // If there's no Date: header, then fall through and return true.
+    if (dateHeader) {
+      var parsedDate = new Date(dateHeader);
+      // If the Date: header was invalid for some reason, parsedDate.getTime()
+      // will return NaN, and the comparison will always be false. That means
+      // that an invalid date will be treated as if the response is fresh.
+      if ((parsedDate.getTime() + maxAgeSeconds) < now) {
```
adityapunjani: Since its "maxAgeSeconds",  it should be `(parsedDate.getTime() + maxAgeSeconds * 1000) < now)`

lib/strategies/cacheOnly.js
```diff
@@ -14,12 +14,21 @@
...
 var helpers = require('../helpers');
 
 function cacheOnly(request, values, options) {
+  options = options || {};
   helpers.debug('Strategy: cache only [' + request.url + ']', options);
   return helpers.openCache(options).then(function(cache) {
-    return cache.match(request);
+    return cache.match(request).then(function(response) {
+      var cacheOptions = options.cache || globalOptions.cache;
+      if (helpers.isResponseFresh(response, cacheOptions.maxAge, Date.now())) {
```
adityapunjani: I am assuming this is `cacheOptions.maxAge` and **not** `cacheOptions.maxAgeSeconds` to be backward compatible with the current behaviour right?

lib/helpers.js
jeffposnick: Good catch!

lib/strategies/cacheOnly.js
adityapunjani: Alright sounds good.

jeffposnick:
Thanks for catching my mistakes, @adityapunjani!

Can you confirm that this change satisfies your use case and resolves #164? I'd like to double-check before merging.

adityapunjani:
@jeffposnick Sure, I will test Flipkart Lite with this branch and get back.

jeffposnick:
Thanks, @adityapunjani! Looking forward to hearing back.

jeffposnick:
I realized that we never merged this.

@adityapunjani, were you able to test out your codebase with this branch to confirm functionality?

adityapunjani:
@jeffposnick Apologies for the delay. We have tested this change in production. However, I noticed  it doesn't work with opaque responses as the date header cannot be accessed. 

We can probably merge this change along with calling out the caveat in the documentation?

adityapunjani:
@jeffposnick Thoughts?

jeffposnick:
Sorry, I lost track of this.

For `opaque` responses, the behavior is going to be the same as it was before this patch, so I don't think there would be any regression. There's nothing better that could be done.

I'll go ahead with fixing the Travis CI build, merging, and then cutting a new release.

PR #209, opened 2016-12-06
Don't fail silently when someone tries to cache a non-GET request

zeorin:
I was unaware (SW n00b) that one could not cache a non-GET request. Unfortunately, `sw-toolbox` has a check for the method of the request and fails silently if it's not a GET request. A simple error in the log would have helped a lot and saved me much time in debugging.

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -21,8 +21,13 @@ var idbCacheExpiration = require('./idb-cache-expiration');
 function debug(message, options) {
   options = options || {};
   var flag = options.debug || globalOptions.debug;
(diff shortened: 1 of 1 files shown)


zeorin:
I signed it!

addyosmani:
Thanks for the PR, @zeorin. My take on this is we should ensure https://github.com/googlechrome/sw-helpers (our next version of this library) includes support for console messages that help folks debug the inability to cache non-GET requests. Would you mind filing an issue upstream for us to discuss and track that there?

zeorin:
Closing due to age.

PR #214, opened 2017-01-11
Test fixes

gauntface:
This should hopefully make the tests happier.

At the very least it bumps everything up to new deps. I moved the npm-publish-scripts back to it's old value because the new version is all CLI based and not module based at all and not sure if it'll ship the right stuff or not :-/


gauntface:
@jeffposnick would you be happy accepting this PR?

Should tide sw-toolbox over for a little while in terms of tests.

jeffposnick:
Cheers! 👍

PR #224, opened 2017-02-16
Publish Sanely

### LINKED ISSUE #223: Investigate publishing issue with 3.5.0 release
I ran into an issue when publishing the `3.5.0` release: the top-level `sw-toolbox.js` and `sw-toolbox.js.map` files weren't copied over from the `build/` directory, and were therefore missing from what got published to `npm`. I published the `3.5.0` release using `npm run publish-release minor`, and under the hood, it uses @gauntface's [`npm-publish-scripts`](https://github.com/GoogleChrome/npm-publish-scripts) project to handle the details. In this particular case, the local [`project/create-release-bundle.sh`](https://github.com/GoogleChrome/sw-toolbox/blob/master/project/create-release-bundle.sh) script was responsible for copying over those files from `build/` to top-level, and it looks like that wasn't automatically run.

I made a local change to copy over the files to top-level and tag `3.5.1` in `package.json`, and then manually ran `npm publish`, so I think it's resolved now.

I'd like to get to the bottom of what happened to ensure that future releases go smoothly. We [currently](https://github.com/GoogleChrome/sw-toolbox/blob/master/package.json#L37) use `npm-publish-scripts` versioned with `^2.0.7`, and looking at my local `node_modules` directory, that was resolved to the actual `2.1.4` release. Taking a look at the [tagged releases](https://github.com/GoogleChrome/npm-publish-scripts/releases?after=v3.0.0-alpha) for `npm-publish-scripts` on GitHub, I don't actually see a final `2.1.4` release there, only a `2.1.4-alpha` and then the `3.0.0-alpha` releases. I wonder if the `2.1.4` release was accidentally published to npm and contains some breaking changes, and that it really should have been tagged 3.0.0?

gauntface:
@jeffposnick apologies for the mess with previous release.

This PR does a few things:

1. Simplifies the release process to just rely on npm's package.json > files to select files to publish.
2. To keep top level sw-toolbox file, changed the build location - not totally happy with this but best solution and I've added files to gitignore (They used to be in master and that's where my dis-taste formed).
3. npm-publish-scripts can be used from the command line and has a reasonable usage flow. Its no longer needed in package.json scripts. I'd still like publishers to use it (It does basic sanity checks like running npm run build, npm run test and creating git tags etc).
4. Publishing docs can now just be done with npm-publish-scripts publish-docs. sw-toolbox has no jsdocs so it's just pushing to github pages.

If you want to confirm that the output to npm is sane, run `npm pack`.

Fixes #223


jeffposnick:
If the idea is that `npm-publish-scripts` is still the recommended way of publishing things, then I'd feel much better if we continued listing it alongside the compatible version info `package.json`, and then ensured that we ran the local copy by referencing it inside of `package.json`'s `scripts` field.

Relying on folks to use a globally installed version of `npm-publish-scripts` without having any means of enforcing versioning seems like it's asking for trouble, and we're run into similar trouble if you released any breaking changes in `npm-publish-scripts` moving forward.

gauntface:
Done.

package.json
```diff
@@ -4,18 +4,16 @@
...
     "test": "gulp lint default && node ./test/helpers/download-browsers.js && mocha",
-    "bundle": "./project/create-release-bundle.sh"
+    "publish-release": "npm-publish-scripts publish-release"
   },
   "main": "lib/sw-toolbox.js",
   "repository": "https://github.com/GoogleChrome/sw-toolbox",
   "dependencies": {
-    "serviceworker-cache-polyfill": "^4.0.0",
-    "path-to-regexp": "^1.0.1"
+    "npm-publish-scripts": "^4.1.0",
```
jeffposnick: This should be in `devDependencies`, I believe.

gauntface: I'm such a noob.

jeffposnick:
Thanks for taking care of this!

PR #232, opened 2017-03-09
Add option to disable errorResponses return success cache

zaaack:
If Http error occurred when success cache exists, the `networkFirst` strategy will always return the success cache. 

I think just return the success cache for errorReponses might be not good for debugging or some other case like the same input cause Http Error in the second time (especially for those who don't want cache much errorResponses and  override `successResponses` option).


zaaack:
I signed it!

gauntface:
@zaaack I'm not sure i follow the use case here.

NetworkFirst is a strategy choice (i.e. you are explicitly wanting to use the cache if the network fails. If that's not what you  want, then you probably want a different strategy or a custom strategy.

zaaack:
@gauntface Here is a use case: somebody made a GET request, succeeded and cached; after a while, the same GET request need permission(like cookie expired) or something else, so the server responds with a 401 error, we need this error to create a message for user, but NetworkFirst strategy returned the old cache, like the request succeed. 

I thought NetworkFirst strategy should always be "network-first", not only successful requests but also requests with an error response.

gauntface:
I'm still lost. If you set that flag on a route, won't you effectively be network only (i.e. the cache will never be used).

zaaack:
Sorry for my late, I haven't using Service Worker for a long time, and this might not be a good way to  solve this problem in the master branch. And since you are already working on a new version in https://github.com/GoogleChrome/workbox (which I haven't try yet), I'll just close this pr, and check workbox for this issue next time I use Service Worker.

PR #240, opened 2017-04-06
Adds support for CacheQueryOptions

### LINKED ISSUE #239: Accept CacheQueryOptions via the `options.cache` config
Now that [`CacheQueryOptions` is widely supported](https://developers.google.com/web/updates/2016/09/cache-query-options), we should let developers pass it in to configure how the underlying `cache.match()` behaves inside a handler.

We're already [doing something similar](https://github.com/GoogleChrome/sw-helpers/blob/master/packages/sw-runtime-caching/src/lib/request-wrapper.js#L55) in `sw-helpers`.

The interface would look something like:

```js
toolbox.router.get(new RegExp('/path/to/file'), toolbox.networkFirst, {
  cache: {
    name: 'my-cache',
    queryOptions: {
      ignoreSearch: true,
      // Any other supported options...
    }
  }
});
```

And then `options.cache.queryOptions` would get passed in as the second parameter to `cache.match()`.

CC: @mohsen1

jeffposnick:
R: @gauntface @addyosmani @wibblymat 

Fixes #239


index.d.ts
```diff
@@ -12,6 +12,7 @@ declare module 'sw-toolbox' {
     name: string
     maxEntries?: number
     maxAgeSeconds?: number
+    queryOptions?: CacheQueryOptions
```
jeffposnick: @mohsen1, can you confirm that this change looks good from a TypeScript perspective? `CacheQueryOptions` is a [standard interface](https://w3c.github.io/ServiceWorker/#dictdef-cachequeryoptions), and I'm not sure if just referencing it like this is sufficient, or whether it needs to be defined as a stub interface inside this file.

mohsen1: Yes this works. DOM tsd files are automatically generated from web idl files.

PR #241, opened 2017-04-06
Remove redundant Request and Response interfaces

mohsen1:
They are defined in [`dom.d.ts`](https://github.com/Microsoft/TypeScript/blob/master/lib/lib.dom.d.ts#L9343-L9374)


jeffposnick:
👍

PR #261, opened 2017-07-27
Support configuring a custom function with the request object as an input param for URL pattern

liuruoran88:
sw-toolbox has two  options, Express-style Routes and 
Regular Expression Routes,  for configuring the routes,  but it's not flexible enough for users. If it can support  configuring a custom function with the request object as an input param for url pattern，users can do more processing according to the request object.  In order to meet this need, I have changed  two files, `lib/router.js` and `lib/route.js`,  and added some test cases in the test folder. These changes will not affect the original Routes config.Your prompt attention to this matter will be appreciated.

An example of using Function Expression routing include:
```js
toolbox.router.get(
function(request){
  // users can get request object, do some processing accroding to the request object and return a boolean value
  if (request && reqest.headers && request.headers.referer) {
     return true;
  } 
}, apiHandler);
```

### DIFF
--- a/lib/route.js
+++ b/lib/route.js
@@ -20,23 +20,23 @@ var url = new URL('./', self.location);
 var basePath = url.pathname;
 var pathRegexp = require('path-to-regexp');
 
-var Route = function(method, path, handler, options) {
-  if (path instanceof RegExp) {
-    this.fullUrlRegExp = path;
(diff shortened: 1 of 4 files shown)


liuruoran88:
I signed it!

liuruoran88:
ok, I have committed another new PR, please check again, thank you!

PR #262, opened 2017-07-27
Support configuring a custom function with the request object as an input param for URL pattern

liuruoran88:
sw-toolbox has two  options, Express-style Routes and 
Regular Expression Routes,  for configuring the routes,  but it's not flexible enough for users. If it can support  configuring a custom function with the request object as an input param for url pattern，users can do more processing according to the request object.  In order to meet this need, I have changed  two files, `lib/router.js` and `lib/route.js`,  and added some test cases in the test folder. These changes will not affect the original Routes config.Your prompt attention to this matter will be appreciated.

An example of using Function Expression routing include:
```js
toolbox.router.get(
function(request){
  // users can get request object, do some processing accroding to the request object and return a boolean value
  if (request && reqest.headers && request.headers.referer) {
     return true;
  } 
}, apiHandler);
```

### DIFF
--- a/lib/route.js
+++ b/lib/route.js
@@ -20,21 +20,23 @@ var url = new URL('./', self.location);
 var basePath = url.pathname;
 var pathRegexp = require('path-to-regexp');
 
-var Route = function(method, path, handler, options) {
-  if (path instanceof RegExp) {
-    this.fullUrlRegExp = path;
+var Route = function(method, pathOrFunc, handler, options) {
+  if (pathOrFunc instanceof RegExp) {
+    this.fullUrlRegExp = pathOrFunc;
+  } else if (typeof pathOrFunc === 'function') {
+    this.funcUrlPattern = pathOrFunc;
   } else {
     // The URL() constructor can't parse express-style routes as they are not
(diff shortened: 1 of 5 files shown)


jeffposnick:
First off, thanks for your contribution!

While `sw-toolbox` continues to be a supported project, we're not looking to make significant changes to the public API at this time.

A newer set of libraries, called Workbox, already provides similar functionality to what you're proposing. Specifically, the [`WorkboxSW.router.registerRoute()`](https://github.com/GoogleChrome/workbox/blob/master/packages/workbox-sw/src/lib/router.js#L92) method will allow you to pass in a custom function for determining whether there's a match, and then apply a handler similar to what's supported in `sw-toolbox`.

If you'd prefer to continue using `sw-toolbox`, what you're proposing is possible without actually changing the public interface if you avoid using the `sw-toolbox` router, and instead use your own custom routing logic alongside the `sw-toolbox` strategies, like:

```js
importScripts('/path/to/sw-toolbox.js');

self.addEventListener('fetch', event => {
  if (event.request.headers.has('referer')) {
     event.respondWith(toolbox.networkFirst(event.request));
  }
});
```

PR #288, opened 2019-04-02
Deprecation notice

jeffposnick:
R: @philipwalton

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,10 @@
-# Service Worker Toolbox
+# Deprecated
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,6 +1,10 @@
-# Service Worker Toolbox
+# Deprecated
```
philipwalton: Suggestion: I think I'd keep the name at the top, but if you want it to stand out more you could always add HTML tags (which I believe work in GitHub markdown files) or use emoji to call attention to it (e.g. ⚠️).

